PR #2, opened 2016-01-15
Add search activities

ylecollen:


ylecollen:
Add the search in roomS activity.

This activity should be improved later by adding the search by room and member names.

PR #8, opened 2016-02-09
Add login splash activity

ylecollen:


ylecollen:
-> First implementation of the login page and the searches activitivies

PR #14, opened 2016-02-17
Merge chat open header

ylecollen:


ylecollen:
Thing to improve
1- dismiss the header when the list is scrolled (MatrixMessageListFragment::onListTouch could warn the activity)
2 - dismiss the header when the keyboard is opened (when the edittext gains the focus)

pedroGitt:
Added the following behaviours:
- room header view is dismissed when: the input text area is touched, when the any bottom boutton is pressed, when the list is scrolled.
- room header view is disabled (cannot be displayed) when the keyboard is displayed
- room header content is fully refreshed with the SDK events

PR #48, opened 2016-03-25
UI/UX improvements

aviraldg:


aviraldg:
For anyone looking at this, https://github.com/vector-im/vector-android/pull/48/commits/4bcaa70ae435600695ba9a7b3f9797bd8a25ee9f is still pretty incomplete and among other things, changes the existing style even for the light theme. A future commit will fix this and revert to the old styling, while only changing it for the dark theme.

ara4n:
@aviraldg - i think we spoke about this at the time, but the reason this hasn't been merged is 'cos of the mix of subjective cosmetic changes (e.g. the colour scheme and animations) and more neutral stuff like the (c) improvements.  I suggest we'll cherrypick from these when we do the next wave of UI re-work on the app.  Thanks tho!

PR #50, opened 2016-03-25
Fix issue  "Rotating screen on VectorHomeActivity loses expander stat…

pedroGitt:
…e. #44" (https://github.com/vector-im/vector-android/issues/44)

- keep track of the group expanding state in the adpater and notify the fragment
- the fragment manages the group expanding state through its life cycle
- fix also some typo issues


ylecollen:
LGTM

PR #79, opened 2016-04-15
General bugfix before delivery#1

pedroGitt:
- fix:  Members with same power levels not allowed to kick each others #77 
- new group expanding management: the expanding state is now aved in the sharedrepository


ylecollen:
LGTM

PR #80, opened 2016-04-15
Bug fixes ylc

ylecollen:


ylecollen:
-> Improve the universal link management
-> application is launched from an universal link :  launch the room activity if the room is known, else wait the next sync
-> login screen : the universal link and the identity server were lost after a screen rotationn

pedroGitt:
LGTM! Go Yannick!!

PR #102, opened 2016-05-02
Update room members list

ylecollen:
The room members list are now sorted with the presence instead of alphabetical order.

So, the criteria are by order
1- power level first
2- presence
3- alphabetical sort.


ylecollen:
LGTM

PR #107, opened 2016-05-03
Add next link logout (registration feature)

pedroGitt:
Adding the registration feature, based on the ownership of an email. Once in the device mailer, from the link in the validation mail, Vector App is started to finish the registration process.


ylecollen:
LGTM

PR #118, opened 2016-05-11
Update default hs

ylecollen:
Vector.im is used now as default home server URL.
The client fallbacks to matrix.org if the login or the forgot password fail.


ylecollen:
LGTM

Issue #125, opened 2016-05-13
If you specify a custom homeserver, the app should remember what it is


matrixbot:
Created by @ matthew:matrix.org.

pedroGitt:
Fixed in: PR Fix home/identity server saving value #135

PR #135, opened 2016-05-18
Fix home/identity server saving value

pedroGitt:
… what it is #125" - the home and identity server URLs are now saved in the SharedPreferences

 - Adding logs to track the "Start chat" action from the VectorMemberDetailsActivity
 - Updating log format for the GCM notifications


ylecollen:
LGTM

PR #141, opened 2016-05-20
Fix landscape photo sent in portrait issue123

pedroGitt:


pedroGitt:
The fix consists, for the MediasPIcker activity, in forcing the screen roation being always managed by the sensor, to by pass any user setting tuning off the sensor-based rotation

PR #150, opened 2016-05-24
Add join room by alias

ylecollen:


ylecollen:
Should fix https://github.com/vector-im/vector-android/issues/136

Issue #198, opened 2016-06-03
"Read All" in Recents page : A spinner should be displayed while sending the read receipts


ylecollen:
By now, there is no spinner so the user could expect that the dedicated read receipts are sent which might be wrong. Thus, the unread counters are not cleared.

PR #202, opened 2016-06-06
Fix defects before delivery

pedroGitt:
Fix Read All" in Recents page : A spinner should be displayed while sending the read receipts #198


ylecollen:
LGTM

PR #214, opened 2016-06-08
Fix defects: #197 Room members : the Pen menu icon should be hidden if the user is alone in the room or is not administrator

pedroGitt:
Fix the defect #197:  Room members : the Pen menu icon should be hidden if the user is alone in the room or is not administrator


ylecollen:
LGTM

PR #215, opened 2016-06-08
Improve medias management

ylecollen:
The application used to support only Uris attachements whereas some attachments could be text (link in a webbrowser) or a contact in xcf format (from the phonebook)


pedroGitt:
LGTM

Issue #192, opened 2016-06-03
"Notification targets" (global settings) entry should not be displayed if it is empty 


ylecollen:

ylecollen:
Should be fixed with
https://github.com/vector-im/vector-android/commit/df1a6ebe1873f60a6f8ec6a45b676b019db3e510

PR #217, opened 2016-06-09
Fix notification targets displayed empty deffect192

pedroGitt:
- fix defect #192 
- fix waiting spinner screen in the vector settings screen (when the waiting screen was displayed, it was still possible access the settings)


ylecollen:
LGTM

PR #218, opened 2016-06-09
Invited detail activity with no avatar

ylecollen:
1- the avatar was not always displayed on details member activity
2 - the members list was not sorted and refreshed when a presence is updated.


ylecollen:
LGTM

Issue #210, opened 2016-06-07
"Start chat" should not be displayed in the member details activity if the room has only 2 members


ylecollen:

ylecollen:
https://github.com/vector-im/vector-android/commit/36c6d50637b15744dc98eede2385535cf24bb257

Issue #186, opened 2016-06-03
Start chat with a member should use the latest room instead of the first found one


ylecollen:
On the detailled member sheet, it is possible to start a chat but many 1:1 chat exists with this user, the one with the most recent message should be used

ylecollen:
Should be fixed with 
https://github.com/vector-im/vector-android/commit/36c6d50637b15744dc98eede2385535cf24bb257

PR #221, opened 2016-06-10
Fix Start chat with a member should use the latest room instead #186

pedroGitt:
- When "Start chat" is pressed (in the member details page) the 1:1 room with the most recent message is chosen in case there are more than 1:1 rooms
- fix some typos, add some logs in catchs, remove some magic number
- add new methods: ```findOneToOneRoomList(), findLatestOneToOneRoom() and isOneToOneRoomJoinedByUserId()``` in CommonActivityUtils.java

Fix:" 'Start chat' should not be displayed in the member details activity if the room has only 2 members #210"
Warning: requires a specific SDk version where Room has the new method getJoinedMembers()


ylecollen:
LGTM

PR #222, opened 2016-06-10
restore_events_polling_thread

ylecollen:
-> should fix https://github.com/vector-im/vector-android/issues/133
The polling thread does not seem to be used when GCM is disabled

-> the events polling thread is restored when GCM is disabled
-> make GA and GCM use configurable (gradle)
-> update build script to build a FDroid version (without GCM) and a google play version (with GCM)


pedroGitt:
- GcmRegistrationManager.java:
 - Why new field is named ```sUseGCM``` and not ```mUseGCM```? even if it is a static field should not it ne named mUseGCM
 - ```useGCM()```: missing code in ```catch()``` section, at least a Log msg..
 - ```setUseGA()```: missing code in ```catch()``` section, at least a Log msg..
 - method naming ```updateListenerNotification()```: since this method sets the service in foreground or remove it from foreground , why it is "only" called updateListener. The important point here is that this method modifies the priority of process the service belongs to.

Apart these remarks LGTM!

PR #229, opened 2016-06-20
Add background sync custo

ylecollen:


ylecollen:
Let the user defines the "sync/" request timeout and the delay between two sync requests

ylecollen:
LGTM

PR #239, opened 2016-06-22
Display notifications when GCM is enabled and background synd is disabled.

ylecollen:
https://github.com/vector-im/vector-android/issues/235
Display notifications when the background is disabled and GCM enabled


ylecollen:
LGTM

Issue #230, opened 2016-06-20
Service is not stopped when user is logged out


pedroGitt:
The Service is not stopped when the user is logged out.
A stopSelf() is missing when EventStreamService is stopped. 
Also, wwhen the user is logged out and Vector is paused (swicth to another App), the PAUSED event must not re start the EventStreamService.

Steps to reproduce:

1. Vector is launched
2. in the slider menu press "Logout"
3. checking the service status (ex: ```adb shell dumpsys activity package im.vector.alpha```) indicates the service is still running

ylecollen:
Should be fixed on
https://github.com/vector-im/vector-android/commit/e617baa5f63c9144d4026efcddd599611c1d67c0

PR #240, opened 2016-06-22
 Fix " Service is not stopped when user is logged out #230"

pedroGitt:
- actions to service are not anymore sent if the user is logged out (ie. no default session available)

 Clean code: typo comments, light renaming, transfer method calls from onCreate() to onResume()


ylecollen:
LGTM

Issue #177, opened 2016-06-03
Click on a room invitation notification should open the room preview


ylecollen:
By now, the room is automatically opened (the VectorRoomActivity is launched)

ylecollen:
https://github.com/vector-im/vector-android/commit/aa28f3329cb5581924261544db3f11f22cf7b884

PR #245, opened 2016-06-23
Fix " Click on a room invitation notification..  #177"

pedroGitt:
Fix  Click on a room invitation notification should open the room preview #177
- in the build notification flow, test if the notification deals with an invitation

### DIFF
--- a/vector/src/main/AndroidManifest.xml
+++ b/vector/src/main/AndroidManifest.xml
@@ -242,6 +242,10 @@
             </intent-filter>
         </activity>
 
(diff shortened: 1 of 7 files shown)


ylecollen:
it does not work on some cases
For example, you're previewing a room and a room invitation is received

pedroGitt:
Just updated the code to delay the construction of the intent that will start the RoomActivity. A fake activity (VectorFakeRoomPreviewActivity) has been added to build the RoomActivity in preview mode, after the user press the "Open" button in the notification of the invitation

ylecollen:
LGTM

Issue #203, opened 2016-06-07
Login page : the buttons should be properly enabled/disabled when the network connection is lost/retrieved 


ylecollen:
When the application is launched in offline mode, the buttons are disabled.

But, the buttons should be enabled when the data connection is retrieved.

Currently, the application must be restarted.

PR #257, opened 2016-06-27
Fix deffect #203 (login page buttons disabled when no network)

pedroGitt:
Deffect: " Login page : the buttons should be properly enabled/disabled when the network connection is lost/retrieved #203"

- when in the login screen if the network is not available, the buttons are enabled as soon as the network is back
- a listener registration on the network status is performed

Renaming: setFlowsMaskEnabled() in enableLoadingScreen()
Fix blank screen: a blank screen was displayed checkNameAvailability() (registration flow) when the network was available
-  onRegistrationEnd() was not called in checkNameAvailability()


ylecollen:
LGTM

PR #266, opened 2016-07-01
Add pending call cells

ylecollen:
Display a cell in the home and the chat activities when there is a pending call.


pedroGitt:
:+1:

PR #277, opened 2016-07-05
Move read contact permission

ylecollen:
-> Unify search 
the read contact permissions is moved from the home activity to the unify search activity when the "People" tab is selected.

-> display a warning message when the user cancels the permissions explanation dialog.


pedroGitt:
:+1:

PR #342, opened 2016-07-26
Make media download upload cancellable

ylecollen:
The medias upload and download can be cancelled.


pedroGitt:
Remarks

1. In "Massages not sent.Resend now?" an extra space between "(..)sent." and "Resend now?" should be added
2. When a download is canceled by the user ("Cancel Upload" menu) and then the user redact ("Redact" menu) the corresponding download, the "Massages not sent.Resend now?" message is not removed
3. When a download is canceled by the user ("Cancel Upload" menu selected), should the message "Massages not sent.Resend now?" be displayed since it is a direct action performed by the user

PR #354, opened 2016-07-28
improve_member_presence_management

ylecollen:
The users are now stored into the application store so it might solve some invalid presences while refreshing them.


pedroGitt:
LGTM :+1:

Issue #343, opened 2016-07-26
Incoming calls should put the application in foreground


ylecollen:
Currently, the client 
-> is ringing
-> triggers an incoming call notification (tapping on it opens the application)

ylecollen:
https://github.com/vector-im/vector-android/commit/9dab4a553ba9a32c0ea974d859706dd1d4b9cdeb

PR #365, opened 2016-08-03
Put app in foreground incoming call #343

ylecollen:
1- any incoming call should put the application in foreground
2- there are 4 types of notification
  -> listening for events
  -> message one
  -> incoming call
  -> call in progress

3- fix some issues


pedroGitt:
- **EventStreamService**
 - Notification IDs line 110: unify naming with a prefix such as ```NOTIF_ID_..``` (ex: ```NOTIF_ID_MESSAGE_NOTIFICATION```)
 - ```onIncomingCall()```: same conditon seems to be tested twice (```state``` vs ```IMXCall.CALL_STATE_CREATED```)
 - ```checkNotification()``` line 1025: add inforation to log (```Log.d(LOG_TAG, "checkNotification " + "SessId="+mNotificationSessionId + " - " + "RoomId="+mNotificationRoomId + " - " + "EventId="+mNotificationEventId);```)
 - ```prepareNotification()``` line 819: log to be updated
- **NotificationUtils**
 - ```buildIncomingCallNotification()``` comment header: add info that this method also starts the VectorHomeActivity which is in charge of centralizing the incoming call flow.
 - ```buildIncomingCallNotification()```: some sanity checks may be missing.. 
- **VectorCallViewActivity**
 - ```getActiveCall()```: add some light comments to explain the code where hideCallNotifications() may be called if the call cannot be resumed..

Issue #383, opened 2016-08-16
Vector should not have the username and password pre-filled in the login boxes after logging out


matrixbot:
Created by @ dave:matrix.org.

PR #391, opened 2016-08-18
Fix login/password kept after logout

pedroGitt:
Fix: Vector should not have the username and password pre-filled in the login boxes after logging out #383


ylecollen:
LGTM

PR #438, opened 2016-09-01
Add contacts access any android

ylecollen:
The contacts book access can be restricted in any android devices.
Android >= M : use the system permission
Android < M : use an application preferences.


ylecollen:
should fix https://github.com/vector-im/vector-android/issues/432 
Warn explicitly that your addressbook's emails are uploaded for discovering matrix IDs if you enable that permission

ylecollen:
LGTM

PR #442, opened 2016-09-01
Fix some other infinite call ringing cases

ylecollen:
https://github.com/vector-im/vector-android/issues/437
Answering a call on desktop whilst also ringing on android causes android to ring forever unstoppably


ylecollen:
LGTM

PR #450, opened 2016-09-04
Swap call and send button to increase UX

andir:
It is alot easier to click on a button on the right edge of the screen
then clicking on a button between the textarea and the voice call
button.

This should help with reducing the amount of unwanted phone calls
started from the mobile app.

### DIFF
--- a/vector/src/main/res/layout/activity_vector_room.xml
+++ b/vector/src/main/res/layout/activity_vector_room.xml
@@ -187,21 +187,7 @@
                         android:textCursorDrawable="@drawable/searches_cursor_background"
                         android:textColorHint="@color/vector_0_54_black_color"/>
 
-                    <RelativeLayout
-                        android:id="@+id/room_send_layout"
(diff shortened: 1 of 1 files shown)


ylecollen:
You should submit your idea to https://vector.im/beta/#/room/#vector-feedback:matrix.org.

We have to keep the same UI on both mobile clients.

The webclient already has the buttons in the order you expect.

andir:
I tried to bring that topic up but seems rather challenging given the traffic in the channel. I could just provide a similar PR for iOS but I can't test nor compile it...

ara4n:
I agree with @andir on this - @ylecollen, @giomfo - let's switch them over on both iOS & Android.

4nd3r:
#395 ?

ylecollen:
The call button is hidden when there is an edited text like 4nd3r said.

ylecollen:
screenshot_2016-09-20-11-09-15
screenshot_2016-09-20-11-09-27

PR #458, opened 2016-09-06
join_conference_voice_video_#454

ylecollen:
Should fix https://github.com/vector-im/vector-android/issues/454
Let users join confs as voice or video


pedroGitt:
:+1:  LGTM

PR #459, opened 2016-09-06
Warning message when encrypted content is received (Fix #455)

### LINKED ISSUE #455: Until e2e is impl'd, encrypted msgs should be shown in the UI as unencryptable warning text
matches vector-im/vector-ios#559

pedroGitt:
- Refactoring code to use a defined constant string for "org.matrix.custom.html"
- Add specific message ("Encrypted message") when an encrypted message (m.room.encrypted) is received.


ylecollen:
LGTM

PR #469, opened 2016-09-08
Fix speaker and mute icons should be available asap.. #431

pedroGitt:
…aunched #431

Add fading effect when in video call


ylecollen:
LGTM

PR #482, opened 2016-09-09
Improve call avatar

ylecollen:
in 1:1 call, display the other call member infos (avatar and displayname) whatever the room infos are.

Should fix
https://github.com/vector-im/vector-android/issues/481
https://github.com/vector-im/vector-android/issues/481


pedroGitt:
- **```VectorUtils```**
 - ```createAvatarThumbnail()``` line 308: magic number - what is the meaning of 42?
 - ```createAvatar()```:``` @param backgroundColor``` is missing
 - ```createAvatar()``` line 329: meaning of the 2 / 3? Should a float constant be defined?
 - ```loadCallAvatar()``` line 494: no sense comment
 - ```loadCallAvatar()``` line 471/481: could not it be simplified to

            ViewParent parent = imageView.getParent();
            while ((pixelsSide < 0) && (null != (parent = imageView.getParent()))) {
                if (parent instanceof View) {
                    View parentAsView = (View) parent;
                    pixelsSide = parentAsView.getLayoutParams().width;
                }
            }
?

pedroGitt:
:+1:

PR #485, opened 2016-09-12
improve_network_connection_detection

ylecollen:
in some cases, the network change receiver is not called when the network connection is updated.

BY now, the SDK will automatically checks the network connection if a request succeeds and the datat connection was marked as "not connected".


pedroGitt:
LGTM

PR #497, opened 2016-09-13
voip_caller_ringtone_#493

ylecollen:
should fix

-> https://github.com/vector-im/vector-android/issues/493
Voip caller : the ringtone should be played in the earspeakers instead of the loud speakers

-> https://github.com/vector-im/vector-android/issues/491
The speaker is turned on when placing a Voice call


pedroGitt:
LGTM :+1:

PR #518, opened 2016-09-16
Fix calls headset issues

ylecollen:
Should fix
https://github.com/vector-im/vector-android/issues/517
Connecting a headset while on speaker should shift audio output to headset


pedroGitt:
:+1:

PR #524, opened 2016-09-17
Changed the UX of the message background from white to gray 

RajPratim21:
can easily distinguish between different messages and also gives a better user view distinguishes the message content from background . it's still incomplete though some more tweaks I am planing to add.
screenshot_2016-09-17-12-03-40


ara4n:
hi @RajPratim21 - you've tried to merge your personal develop branch onto vector's master branch (you should be targeting vector's develop branch), and you also don't seem to have pushed any commits to your personal develop branch.

In terms of a custom theme: please feel free to customise as desired, but we're not going to accept any PRs to the main application for changes to the UI based on subjective personal taste.  In future we may support customisable skins in the app, in which case alternative skin proposals would be much appreciated.

Issue #531, opened 2016-09-20
The application badge should be updated even if the device is offline.


ylecollen:

ylecollen:
BY now, the badge value is the number of rooms which has unread notified messages.
This value is updated by the push notifications.

But when the device is offline or the application is the fdroid one, the client should be able to manage the badge.

PR #537, opened 2016-09-20
Fix The application badge should be updated even if the device is off… #531

pedroGitt:
Fix the badge update when the device is offline
Update of the SDK lib with its last version


ylecollen:
The invited rooms are not counted to update the badge (need to check if the dedicated rule is enabled / disabled)

vector/src/main/java/im/vector/activity/VectorHomeActivity.java
```diff
@@ -774,6 +779,47 @@ private void markAllMessagesAsRead() {
...
+     * message of the room.
+     * @return true if operation was performed, false otherwise
+     */
+    private boolean markAllMessagesAsReadWhenOffline() {
+        boolean isOperationDone = false;
+
+        if(!Matrix.getInstance(this).isConnected()) {
+
+            ArrayList<MXSession> sessionsList = new ArrayList<>(Matrix.getMXSessions(this));
+            mReadReceiptSessionListIterator = sessionsList.iterator();
```
ylecollen: it seems being a copy & paste from the standard case.
It seems useless here.

vector/src/main/java/im/vector/activity/VectorHomeActivity.java
```diff
@@ -774,6 +779,47 @@ private void markAllMessagesAsRead() {
...
+                for (MXSession session : sessionsList) {
+                    if (null != session) {
+                        // test if the session is still alive i.e the account has not been logged out
+                        ArrayList<Room> roomCompleteList = new ArrayList<>(session.getDataHandler().getStore().getRooms());
+
+                        if(null != roomCompleteList) {
+                            // for each room send the receipt for the latest message
+                            for (Room room : roomCompleteList) {
+                                room.sendReadReceipt(null);
+                                isOperationDone = true;
```
ylecollen: isOperationDone |= room.sendReadReceipt(null);
seems better : it is possible there is no update. sendReadReceipt returns true when there is an update.

vector/src/main/java/im/vector/activity/CommonActivityUtils.java
```diff
@@ -1684,6 +1685,50 @@ public static int getBadgeCount() {
...
     }
 
+    /**
+     * Refresh the badge count when the device is offline.
+     * Notifications rooms are parsed to track the notification count value.
+     * @param aSession session value
+     * @param aActivity calling activity
+     */
+    public static void offlineRefreshBadgeUnreadCount(MXSession aSession, Activity aActivity) {
+        MXDataHandler dataHandler;
```
ylecollen: i don't understand why an Activity is required here.
A context seems better.
For the linked f-droid issue, an activity won't be always available.

pedroGitt: ACK

vector/src/main/java/im/vector/activity/VectorHomeActivity.java
pedroGitt: ACK

vector/src/main/java/im/vector/activity/VectorHomeActivity.java
pedroGitt: ACK

PR #542, opened 2016-09-21
Fix The camera permission should be not required in the room settings…

pedroGitt:
… if the user cannot update the room avatar. #532

- Before asking the camera permission, the user power level is checked against level required to update the avatar


ylecollen:
LGTM

Issue #503, opened 2016-09-14
register users without email verification 


farenhite:
vector is very good.please add ability to users can register without email verification on custom servers

ara4n:
Email registration should always be optional?  Just leave the email 
address blank?

farenhite:
i mean from android client not from web. from web yes its true i can leave the email blank and its ok but when i try to register from android client and leave it blank i get error on my custom server but when home server is matrix.org when i leave email blank it go to capture section and then ok whats problem? problem is my home server config?so what can i do

ylecollen:
As you said, it works properly with matrix.org HS.

Which error message  do you have ?
I assume there is something wrong in the android client registration flow with your HS.

Could you provide us your HS URL ? we could test (and fix) where the issue is.

farenhite:
my HS url = https://104.237.225.220:8448

and my android studio log and error is:

09-15 02:57:17.164 2287-2424/im.vector.alpha D/Retrofit: <--- HTTP 400 https://104.237.225.220:8448/_matrix/client/r0/register (946ms)
09-15 02:57:17.164 2287-2424/im.vector.alpha D/Retrofit: : HTTP/1.1 400 Bad Request
09-15 02:57:17.164 2287-2424/im.vector.alpha D/Retrofit: Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept
09-15 02:57:17.164 2287-2424/im.vector.alpha D/Retrofit: Access-Control-Allow-Methods: GET, POST, PUT, DELETE, OPTIONS
09-15 02:57:17.164 2287-2424/im.vector.alpha D/Retrofit: Access-Control-Allow-Origin: *
09-15 02:57:17.164 2287-2424/im.vector.alpha D/Retrofit: Content-Type: application/json
09-15 02:57:17.164 2287-2424/im.vector.alpha D/Retrofit: Date: Thu, 15 Sep 2016 06:57:17 GMT
09-15 02:57:17.164 2287-2424/im.vector.alpha D/Retrofit: OkHttp-Received-Millis: 1473922637163
09-15 02:57:17.164 2287-2424/im.vector.alpha D/Retrofit: OkHttp-Response-Source: NETWORK 400
09-15 02:57:17.164 2287-2424/im.vector.alpha D/Retrofit: OkHttp-Selected-Protocol: http/1.1
[...]

PR #550, opened 2016-09-23
Register users without email verification #503

ylecollen:
should fix issue #503 
register users without email verification


pedroGitt:
LGTM

Issue #535, opened 2016-09-20
Fdroid : the application badge value is not supported


ylecollen:
The application badge value defines the number of rooms with unread notified messages.

The badge is updated by the GCM notifications. So, with the F-droid client, this counter is never updated.

It would be nice to have the same feature than the google play version.

Matrixcoffee:
#446

PR #559, opened 2016-09-26
Fix Fdroid : the application badge value is not supported enhancement…

pedroGitt:
… F-Droid feature #535

- For device not supporting GCM, the badge count is now updated
- the badge count is computed for every LiveEventsChunkProcessed notification

### DIFF
--- a/vector/src/main/java/im/vector/services/EventStreamService.java
+++ b/vector/src/main/java/im/vector/services/EventStreamService.java
@@ -282,13 +282,6 @@ public class EventStreamService extends Service {
             triggerPreparedNotification(true);
             mPendingNotifications.clear();
 
-            // update badge unread count in case device does not support GCM (ie. FDroid)
-            MXSession defaultSession;
-            Context appContext = getApplicationContext();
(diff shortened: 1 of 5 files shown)


vector/src/main/java/im/vector/services/EventStreamService.java
```diff
@@ -281,7 +281,14 @@ public void onBingEvent(Event event, RoomState roomState, BingRule bingRule) {
...
             triggerPreparedNotification(true);
             mPendingNotifications.clear();
-            
+
+            // update badge unread count in case device does not support GCM (ie. FDroid)
+            MXSession defaultSession;
+            Context appContext = getApplicationContext();
+            if((null != appContext) && (null != (defaultSession=Matrix.getInstance(appContext).getDefaultSession()))) {
+                CommonActivityUtils.specificUpdateBadgeUnreadCount(defaultSession, appContext);
+            }
```
ylecollen: Detect if there is a new message, a self read receipt or similar.

ylecollen:
LGTM

PR #592, opened 2016-10-03
Improve members list display

ylecollen:
The room members list was refreshed on UI thread. So, it used to add lags when the room has a lot of members (like Matrix HQ).


ylecollen:
LGTM

PR #606, opened 2016-10-06
Add historical rooms support

ylecollen:
The historical rooms support

-> tap on the header to trigger an initial sync. Display a spinner while syncing them.

-> they are not saved in the application store. They are saved in a memory store.
The historical rooms won't be restored at next application launch.

-> if the left rooms have been synced one, they will be synced like other joined room

->tap on a room will open the preview mode to join again the room

-> the search works on left rooms.


pedroGitt:
- **```VectorRoomSummaryAdapter```**
 - ```getChildView()``` line713:  why not using ```getChildType``` like this ```if (ROW_TYPE_SPINNER == getChildType(groupPosition,childPosition)) {``` ? (more easy to understand)

Issue #634, opened 2016-10-17
Crash on searching for a room


jeangb:
To reproduce:
1. Log-in Riot
2. Hit the search button
3. Enter a text : "test", don't press enter
4. Touch Browse directory
5. Riot crash

Version 0.5.3-dev (8f034c8)

Zarokka:
#641

PR #641, opened 2016-10-20
fix crash on searching a room #634

Zarokka:
- changed room count from Integer to int (adapted related code)
- minor refactoring

I could be wrong, but I think that the room count can be an int (instead of Integer).
This way it does not have to be null checked and can be used more safely. 
I also couldn't quite resist and did some minor refactoring where I changed code.

If this is too much of a change for this small issue, you could just do this instead:

```
/**
* @return the matched public rooms count
*/
public int getMatchedPublicRoomsCount() {
    return null == mMatchedPublicRoomsCount ? 0 : mMatchedPublicRoomsCount.intValue();
}
```

### DIFF
--- a/vector/src/main/java/im/vector/PublicRoomsManager.java
+++ b/vector/src/main/java/im/vector/PublicRoomsManager.java
@@ -36,12 +36,13 @@ public class PublicRoomsManager {
     private static final String LOG_TAG = "PublicRoomsManager";
 
     public static final int PUBLIC_ROOMS_LIMIT = 20;
+    public static final int PUBLIC_ROOMS_NOT_INITIALIZED = -1;
 
     public interface PublicRoomsManagerListener {
         /**
          * Called when the number of public rooms count have been updated
          */
-        void onPublicRoomsCountRefresh(Integer publicRoomsCount);
+        void onPublicRoomsCountRefresh(int publicRoomsCount);
     }
 
     // session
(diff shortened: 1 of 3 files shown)


ylecollen:
Your PR does not really fix the issue : we should not open the public rooms list until it is retrieved.

Zarokka:
Not quite sure what you mean.

The changes are not made where the actual crash happened, maybe that is the problem.
The crash happened in VectorSearchRoomsListFragment.java:140 `mAdapter.getMatchedPublicRoomsCount() > 0` because of the auto unboxing of a null Integer.

Here the full code that prevents the opening of the room list, if there are no rooms (unchanged by this PR):
```
if (TextUtils.isEmpty(mAdapter.getSearchedPattern()) || (mAdapter.getMatchedPublicRoomsCount() > 0)) {
    Intent intent = new Intent(getActivity(), VectorPublicRoomsActivity.class);
    intent.putExtra(VectorPublicRoomsActivity.EXTRA_MATRIX_ID, mSession.getMyUserId());

    if (!TextUtils.isEmpty(mAdapter.getSearchedPattern())) {
        intent.putExtra(VectorPublicRoomsActivity.EXTRA_SEARCHED_PATTERN, mAdapter.getSearchedPattern());
    }
    getActivity().startActivity(intent);
}
```

A null-check could be added there instead.
I just try to fix it more solidly, because a null-check of a count is quite easy to forget and there didn't seemed to be a good reason why this could not be a primitive int.

pedroGitt:
=> #690 
Thanks.

Issue #618, opened 2016-10-11
Implementing remaining 1:1 semantics across the UI 


ylecollen:
We have #2070 for tracking whether given rooms are 1:1 or not.
We have #2097 for showing this in the UI.
We have #1882 for new UI for creating rooms as 1:1.

What we don't have is a bug to track all the other niggling bits in the UI where we want to reflect the new 1:1 semantics:

 The ability to untag a 1:1 room and make it a normal room. This could be drop & drag between sublists in LeftPanel, and probably also a checkbox in RoomSettings.
 For symmetry we should probably have the ability to tag a room as a official 1:1 for a given person. This is more fiddly, and we can probably ignore this for now.
 In MemberInfo, rather than having the "Start Direct Chat" or "Go to Direct Chat" buttons, we should just list the rooms we have in common with a user - first the DMs, then a "Start direct chat" button (which has no reuse semantics), and then any other rooms in common. This is basically #1477. This makes it the user's problem to decide which channel to use when talking to someone.
 When sending an invite, we need to explicitly tag it as DM or not DM. (This is either from the new 'find person' UI in #1882, or from the invite UI - either the legacy MemberList based inviting, or the new style as per #1758)
 When receiving an invite, we need to check if it's tagged as DM or not DM. If it's a DM, we add it to m.direct on account_data. If it has no tag at all, we'll have to apply heuristics to decide whether to consider it a DM (given it may have come from a client that doesn't know about m.direct).
 Ensure that 1:1 rooms are created with equal ops on both side
In #1882, we have the problem of whether the 'People' button should reuse DMs, prompt a selection of DMs to chose from, or start a new one. I'll file that there. (Conclusion is to to reuse if possible for now; in future prompt a selection of DMs and rooms in common to use).

giomfo:
TODO:
- tag as direct a room created with only one participant by using 'start chat' option of the (+) button in main screen.
- remove the section name "People", the direct rooms are dispatched in the existing sections.

giomfo:
More important is actually using the new DM semantics:
 * [ ] When I choose 'start chat' (currently 'invite people') from the plus button: if i only select one person, it should invite them as a DM (i.e. with `is_direct` true).
 * [x] Ability to untag a room as a DM
 * [x] Ability to tag a room as a DM for a given user
 * [x] DMs are tracked in the `account_data`. 
 * [ ] If there is already a DM for the user, we should reuse it when hitting 'start chat'.
[...]

PR #668, opened 2016-11-03
Fix Implementing remaining 1:1 semantics across the UI #618

pedroGitt:
- remove PEOPLE section
- update direct chat icon position
 - if a direct chat already exists with a given participant, it must be reused when "start chat" action is performed on the same participant
 - when a "start chat" action is performed with only one participant, it will be tagged as direct message (is_direct=true)


ylecollen:
LGTM

PR #691, opened 2016-11-09
Add devices verification

ylecollen:


pedroGitt:
**```encrypted_event_info.xml```**

- line 20 ```"Event information"``` title should be defined as a string resource in ```string.xml``` (ex: ```"@string/encryption_information_title"```)

PR #827, opened 2017-01-03
Add 'Call ringtone' setting to change ringtone for incoming calls.

juanitobananas:
Okay, second try. The comment from my first PR (#826) still applies (well, most of it):

Add 'Notification Sounds' category in settings with a 'Call
ringtone' setting and a standard ringtone picker to change it.

I am not sure if you even want to have something like this. I think it is pretty cool :wink: (and pretty common). IMHO the current ring.ogg is a little annoying. I haven't (yet?) deleted the original code.

I hope you like it. Feel free to ignore this if you don't :smile:

I am not sure if creating a RingtoneUtils was a good idea, as part of the code is still VectorCallSoundManager but I find it easier to handle like this.

I will gladly change some parts/style if you want me too. Just say so!

Cheers

### DIFF
--- a/vector/src/main/res/xml/vector_settings_preferences.xml
+++ b/vector/src/main/res/xml/vector_settings_preferences.xml
@@ -77,6 +77,16 @@
 
     <im.vector.preference.VectorDividerCategory />
 
+    <PreferenceCategory android:title="@string/notification_sounds_settings">
+
+        <im.vector.preference.VectorCustomActionEditTextPreference
+            android:key="@string/notification_sounds_settings_ringtone"
+            android:title="@string/notification_sounds_settings_ringtone"/>
+
+    </PreferenceCategory>
+
+    <im.vector.preference.VectorDividerCategory />
+
     <PreferenceCategory
         android:title="@string/settings_notifications_targets"
         android:key="@string/settings_notifications_targets">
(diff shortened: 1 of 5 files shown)


dva-re:
pls, don`t quit this idea.

vector/src/main/res/xml/vector_settings_preferences.xml
```diff
@@ -77,6 +77,16 @@
 
     <im.vector.preference.VectorDividerCategory />
 
+    <PreferenceCategory android:title="@string/notification_sounds_settings">
+
+        <im.vector.preference.VectorCustomActionEditTextPreference
+            android:key="@string/notification_sounds_settings_ringtone"
```
bmarty: key cannot be localized string (it has been fixed for the other preferences)

juanitobananas:
Hi @bmarty,

No problem, thanks for the (late) response! Of course, I'll try to rebase it. I am currently playing around with kotlin, so that might just work :wink: 

I don't think I'll find time for this until September though. But I'd be glad to do it. If you happen to be in a hurry, I am afraid you'll have to do it yourself or find somebody else.

Cheers,
Juanito

bmarty:
Hi @juanitobananas ,

Ok, there is no rush :)

Benoît

bmarty:
Replaced by #2730

bmarty:
Hello @juanitobananas ,
No problem, we apologize for a such long delay to deal with your interesting PR :)
Benoit

sl5net:
ring.ogg  is still ringing at my android. i tried change all sound but ists still rining. and i deleted ring.ogg from the file system (dosent help)

PR #834, opened 2017-01-06
Show incoming call activity over the lock screen

juanitobananas:
I think this would be pretty cool and would make it more similar to how other telephony apps work.

Cheers,
Juanito

### DIFF
--- a/vector/src/main/java/im/vector/activity/InComingCallActivity.java
+++ b/vector/src/main/java/im/vector/activity/InComingCallActivity.java
@@ -24,6 +24,7 @@ import android.text.TextUtils;
 import org.matrix.androidsdk.util.Log;
 import android.view.View;
 import android.view.Window;
+import android.view.WindowManager;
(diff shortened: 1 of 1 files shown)


juanitobananas:
Of course, this could lead to picking calls without actually wanting too. A swipe thingie to pick up would be nice together with this.

ylecollen:
This update must be set by a global settings. I don't want that someone complains that the call has be accepted whereas his phone was in his pocket.

Which android version did you test ?

juanitobananas:
> This update must be set by a global settings. I don't want that someone complains that the call has be accepted whereas his phone was in his pocket.

Of course, that shouldn't be a big deal.

> Which android version did you test ?

I tested it with CM13 (roughly similar to Android 6.0.1).

juanitobananas:
I just tested it with an Android 5 and it doesn't seem to be working so I'll take a look at it and open a new PR if I find a solution to this.

PR #835, opened 2017-01-06
Android automatic refactor - Recycle

luiscruz:
Hi,

I am developing a tool to automatically refactor Android applications with the goal of improving energy efficiency.
This pull request has the changes generated while applying the rule "Recycle".

Some resources (e.g., ```Cursor``` instances) should be closed when they are no longer necessary. 

I have made a previous validation of the changes and they seem correct.
Please consider them and let me know if you agree with them.

Best,
Luis


ylecollen:
Do you really think  it would save battery to call recycle ?
TypedArray is not a large object.

spacekitteh:
@ylecollen it's still best to close resources as soon as you're done with them, though.

PR #836, opened 2017-01-06
android automatic refactor obsoletelayoutparam

luiscruz:
Hi (again),

I am developing a tool to automatically refactor Android applications with the goal of improving energy efficiency.
This pull request has the changes generated while applying the rule "ObsoleteLayoutParam".

While developing your application's views you might be specifying attributes in a view's artefact that are not necessary due to the nature of its parent. In this PR, those attributes were replaced by a comment.

I have made a previous validation of the changes and they seem correct.
Unfortunately, this tool is not able keep the original whitespace of the files, so comparison without ignoring whitespace might be confusing. I've found helpful using ```git diff -w```.

Please consider the changes and let me know if you agree with them.

Best,
Luis


ylecollen:
Did you test your update ?

bmarty:
activity_vector_room.xml has changed too much

PR #845, opened 2017-01-13
Improve people invite screen

sabrinaj:
Replace "no result" by bubble background when search input is empty
Display search input even if not valid email/matrix user id

### DIFF
--- a/vector/src/main/java/im/vector/adapters/ParticipantAdapterItem.java
+++ b/vector/src/main/java/im/vector/adapters/ParticipantAdapterItem.java
@@ -15,6 +15,7 @@
  */
 
 package im.vector.adapters;
+
 import android.content.Context;
 import android.graphics.Bitmap;
 import android.text.TextUtils;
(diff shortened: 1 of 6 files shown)


vector/src/main/java/im/vector/adapters/ParticipantAdapterItem.java
```diff
@@ -67,6 +71,7 @@ public ParticipantAdapterItem(RoomMember member) {
         mDisplayName = member.getName();
         mAvatarUrl = member.avatarUrl;
         mUserId = member.getUserId();
+        mIsValid = true;
```
ylecollen: mIsValid could be globally initilized

vector/src/main/java/im/vector/adapters/VectorParticipantsAdapter.java
```diff
@@ -816,6 +817,9 @@ public void onSuccess(Void info) {
             matrixUserBadge.setVisibility(View.GONE);
         }
 
+        // Add alpha if cannot be invited
+        convertView.setAlpha(participant.mIsValid ? 1f : 0.5f);
+
```
ylecollen: convertView.setEnabled(participant.mIsValid)  should avoid triggering the setOnChildClickListener.onclick event.

PR #848, opened 2017-01-13
Unified search: Display search input even if not valid email/matrix u…

sabrinaj:
Unified search: Display search input even if not valid email/matrix user id


ylecollen:
LGTM

PR #873, opened 2017-01-20
Increase height of the message text box tappable area to full height

vurpo:
The message text box in the app currently only covers about a third vertically of the bar at the bottom of the screen, which makes it easy to miss and not activate it when you tap it. This PR changes the height so the text box covers the entire height of the bar, which makes it activate when you tap anywhere (vertically, not horizontally) inside the bottom bar.

This change doesn't affect the way the text box looks or behaves (cosmetically or otherwise) at all, as far as I have tested it. The change is not a visual change, it only increases the height of the tappable area of the textbox.


ylecollen:
We implemented the expected design.
I will report your remarks to the designers.

PR #880, opened 2017-01-23
Fix switch style in settings/room settings for API < 21

sabrinaj:


vector/src/main/java/im/vector/preference/VectorSwitchPreference.java
```diff
@@ -16,22 +16,26 @@
...
+import im.vector.R;
+
 /**
  * This class should fix the android issue
  * https://code.google.com/p/android/issues/detail?id=26194
  *
  */
-public class VectorSwitchPreference extends SwitchPreference {
+public class VectorSwitchPreference extends CheckBoxPreference {
     private static final String LOG_TAG = "VectorSwitchPreference";
```
ylecollen: VectorSwitchPreference should be renamed VectorCheckBoxPreference

ylecollen:
1 - All the xml preference files must be updated to use CheckBoxPreference instead SwitchPreference.
 You did not update the room settings preference.

2 - it is more a patch than a proper fix : if a newer android version uses another way to display the preferences, it won't do the job.
preferencefragmentcompat would be safer.

PR #892, opened 2017-01-26
Add phonenumbers tools

ylecollen:
Add the phone numbers lookup 
The contacts phone numbers are now converted to E164 format before retrieving their linked Matrix id.


vector/src/main/java/im/vector/util/CountryPhoneData.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public String getCountryCode() {
+        return mCountryCode;
+    }
+
+    public String getCountryName() {
+        return mCountryName;
+    }
+
+    public int getIndicator() {
```
ylecollen: You should describe what Indicator means.
It seems also weird that it is an int.

Issue #904, opened 2017-02-01
Room details: spinning wheel takes age to goes off on FILES layout


jeangb:
To reproduce:

1. Open room details of a room with some files.
2. Hit on the FILES tab

Even if there isn't any files in the room, the calc with spinning wheel takes age to be cleared off.
I don't reproduce it on the web and iOS clients.
My device is a Samsung Galaxy S4.
![spinning wheel on room details file](https://cloud.githubusercontent.com/assets/22364949/22504533/2cdbf492-e877-11e6-922d-be6adcf43f84.png)

jeangb:
@ylecollen KO for rooms with no files, there is still the infinite spinning wheel.
Ok for the others.

jeangb:
Validated in 0.6.9-dev (6336758)

PR #912, opened 2017-02-02
Fix files search mask

ylecollen:
Should fix https://github.com/vector-im/riot-android/issues/904
Room details: spinning wheel takes age to goes off on FILES layout 

Fix the following issues
-> Listing room files fragment : 
  -- there were two spinners in some cases
  -- the loading mask was not removed at the end of the search

-> room search activity 
 -- there was a back pagination spinner in some cases
 -- trigger a search with the same pattern used to display a loading mask which was not removed

--> unified search
 -- the files or messages did not always fill the screen history. The back pagination was broken in this case.

--> room activity
 -- disable the search if the room is encrypted. The server cannot search in an encrypted room.
     Assume that the rooms are created and encrypted asap i.e the room did not live unencrypted before switching.

--> room files list
 -- fix the encrypted room case.


vector/src/main/java/im/vector/fragments/VectorSearchMessagesListFragment.java
```diff
@@ -173,6 +175,17 @@ protected boolean allowSearch(String pattern) {
         return !TextUtils.isEmpty(pattern);
     }
 
+    @Override
+    public void onInitialMessagesLoaded() {
+        // ensure that the list don't try to fill itself
+        // if the search is not allowed with the provided pattern.
+        if (!allowSearch(mPattern)) {
+            android.util.Log.e(LOG_TAG, "## onInitialMessagesLoaded() : history filling is cancelled");
```
sabrinaj: Utiliser l'objet Log déjà importé (org.matrix.androidsdk.util.Log)

Issue #923, opened 2017-02-08
local contact section should be collapsable even when no search is started


giomfo:
According to @AmandineLP:
- the known contacts header should be displayed when no search is started too, with the label: "Known contacts (-)" and the following message: "Too many contacts, please use the search field"
- During the search, the known contacts header should be displayed even if the section is empty: "Known contacts (0)"

jeangb:
Validated and covered by automated test.

PR #925, opened 2017-02-09
Issue #923: local contact section should be collapsible even when no search is started

sabrinaj:


sabrinaj:
See issue https://github.com/vector-im/riot-android/issues/923

Issue #933, opened 2017-02-10
Search from recents: strange behaviour in the differents tab


jeangb:
To reproduce:

1. From the recents, hit the search button, then MESSAGES tab
2. In the search bar, enter a string you know it exists in a room
3. Once the result is found, hit PEOPLE tab, then MESSAGES

Problem 1: 
Page is blank, the previously found result isn't show anymore.
If you erase (X button) the search bar and launch a search with the same string, the page is still blank.
![2](https://cloud.githubusercontent.com/assets/22364949/22830824/38fd383e-efa8-11e6-89a4-98308a7fdb54.jpeg)

4. Hit the ROOMS tab

Problem 2:
The layout which is supposed to be displayed in the MESSAGES tab is displayed in the ROOMS tab:
![1](https://cloud.githubusercontent.com/assets/22364949/22830976/d8c32040-efa8-11e6-93ed-916f92720abb.jpeg)

Doesn't happen all the time.

jeangb:
Validated in 0.6.9-dev (7e0ac6b)

Issue #934, opened 2017-02-10
Search from recents: no results displayed if device is turned landscape then portrait


jeangb:
To reproduce:

1. From the recents lists, hit the search button
2. Hit the MESSAGES tab then launch a search with a string that matches a message in a room
3. Once result(s) is found, turn the device in landscape mode, then portrait again

"No results" is displayed.

jeangb:
Covered by test mobilestests_android.RiotSearchTests.searchRoomsAndMessages()
And validated in 0.6.9-dev (7e0ac6b)

PR #942, opened 2017-02-14
Fix issues #933 + #934

sabrinaj:


ylecollen:
LGTM (after adding the missing copyright lines)

PR #958, opened 2017-02-20
Enable building with Jack

spacekitteh:
jackInProcess is set to false, as I'm pretty sure this allows it to attach to an existing Jack server.

### DIFF
--- a/vector/build.gradle
+++ b/vector/build.gradle
@@ -17,6 +17,11 @@ android {
         // use the version code
         versionCode rootProject.ext.versionCodeProp
         versionName rootProject.ext.versionNameProp
+        
+        jackOptions {
+            enabled true
+            jackInProcess false
+        }
     }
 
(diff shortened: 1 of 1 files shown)


ylecollen:
http://tools.android.com/tech-docs/jackandjill 
Jack build chain is still experimental so i'm not sure that it is safe to use it.

Did you already use it in other projects ?

spacekitteh:
Yes, it is the default build chain for AOSP 7.0 and up. :)

ylecollen:
our target SDK is still 23 so i prefer to keep the default build chain until we support the target API becomes 24/

jcgruenhage:
Actually, JACK has been deprecated, in case you didn't notice, so switching to JACK now wouldn't do any good.

bmarty:
Jack is deprecated, I close this PR

PR #965, opened 2017-02-22
Delete 3pid feature

sabrinaj:
+ Delete 3pid feature
+ Add "done" button to keyboard on phone verification screen


vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.java
```diff
@@ -1358,8 +1419,10 @@ private void refreshPushersList() {
      * Refresh the emails list
      */
     private void refreshEmailsList() {
+        final List<ThirdPartyIdentifier> currentEmail3PID = mSession.getMyUser().getlinkedEmails();
+
```
ylecollen: You should create a copy of the list (it can be either in the SDK or here).
It avoids crash when the list is updated while enumerating it).

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.java
```diff
@@ -1565,8 +1635,10 @@ public void onClick(DialogInterface dialog, int which) {
      * Refresh phone number list
      */
     private void refreshPhoneNumbersList() {
+        final List<ThirdPartyIdentifier> currentPhoneNumber3PID = mSession.getMyUser().getlinkedPhoneNumbers();
+
```
ylecollen: Same comment than the email.

PR #971, opened 2017-02-23
Fix incoming call with unknown devices

ylecollen:
Display an unknown devices dialog when an incoming call is received in a room with unknown devices.

Add the "send anyway" button to resend the messages from the dialog.


vector/src/main/java/im/vector/activity/InComingCallActivity.java
```diff
@@ -40,7 +44,7 @@
  * InComingCallActivity is Dialog Activity, displayed when an incoming call (audio or a video) over IP
  * is received by the user. The user is asked to accept or ignore.
  */
-public class InComingCallActivity extends Activity { // do NOT extend from UC*Activity, we do not want to login on this screen!
+public class InComingCallActivity extends AppCompatActivity {
```
sabrinaj: Vector copyright is missing

PR #838, opened 2017-01-07
Start riot automatically on boot

juanitobananas:
If not, you can't receive any messages until you open the app manually.

Unfortunately, you need a new permission for this.

Cheers,
Juanito

PR #1003, opened 2017-03-08
Start on boot if background sync is allowed and GCM is not supported

### LINKED ISSUE #319: Listen for events on boot
Firstly, thanks for your work.

Please, could you add an option to start to listen events on boot?

Thank you.

Unrud:
Basically the same as #838 but only for the FDroid flavor and with some additional checks. (Fixes #319)

### DIFF
--- a/vector/src/appfdroid/AndroidManifest.xml
+++ b/vector/src/appfdroid/AndroidManifest.xml
@@ -1,16 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="im.vector" xmlns:tools="http://schemas.android.com/tools" android:installLocation="auto">
-
-    <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
-
-    <application>
-        <receiver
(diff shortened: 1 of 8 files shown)


ylecollen:
It might work only for the Fdroid clients.
Your implementation does not check if the installed app is the Google play or the Fdroid one.

Unrud:
> Your implementation does not check if the installed app is the Google play or the Fdroid one.

All changes are in the folder *vector/src/**appfdroid**/*.

ylecollen:
Sorry, i did not notice it.
But, i think it should be optional : we need to have a settings entry to let the user to decide.

Unrud:
I moved the code from **appfroid** to **main** and instead added a check for GCM support.

> But, i think it should be optional : we need to have a settings entry to let the user to decide.

Done. (Although I think the option is superfluous, you can already disable background sync.)

ylecollen:
It does not work on a nexus 5X (android 7.1)

ylecollen:
i rewrote a part of your code because the develop branch has been updated a lot since this PR.
https://github.com/vector-im/riot-android/commit/2ea7205195de12899572766d5d401294e0598fc1

Issue #1004, opened 2017-03-08
Crash on the contact picker


jeangb:
To reproduce:
1. Hit + button, then Start chat

Or

1. Hit the search button from recents, then people tab

Riot crashes

ylecollen:
03-08 18:44:24.037 24434 24434 E AndroidRuntime: FATAL EXCEPTION: main
03-08 18:44:24.037 24434 24434 E AndroidRuntime: Process: im.vector.alpha, PID: 24434
03-08 18:44:24.037 24434 24434 E AndroidRuntime: java.lang.NullPointerException
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at java.util.regex.Matcher.reset(Matcher.java:177)
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at java.util.regex.Matcher.<init>(Matcher.java:90)
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at java.util.regex.Pattern.matcher(Pattern.java:297)
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at im.vector.contacts.PIDsRetriever.retrieveMatrixIds(PIDsRetriever.java:201)
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at im.vector.contacts.ContactsManager$4.run(ContactsManager.java:393)
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at android.os.Handler.handleCallback(Handler.java:733)
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:95)
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:157)
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:5356)
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at java.lang.reflect.Method.invokeNative(Native Method)
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Method.java:515)
03-08 18:44:24.037 24434 24434 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit

ylecollen:
fixed on the develop version.

jeangb:
Validated

PR #1008, opened 2017-03-09
Fix msisdn lookup

sabrinaj:


ylecollen:
fix https://github.com/vector-im/riot-android/issues/1004

PR #1009, opened 2017-03-09
Registration with phone number

sabrinaj:


vector/src/main/java/im/vector/RegistrationManager.java
```diff
@@ -0,0 +1,925 @@
...
+     * Request a validation token for the given three pid
+     *
+     * @param pid
+     * @param listener
+     */
+    private void requestValidationToken(final ThreePid pid, final ThreePidRequestListener listener) {
+        if (getThirdPidRestClient() != null) {
+            switch (pid.medium) {
+                case ThreePid.MEDIUM_EMAIL:
+                    String webAppUrl = "https://vector.im/develop";
```
ylecollen: Why this URL is hard coded ?
it should be riot.im/app or so ?

ylecollen: It would be nice to add a constant value at the top of this class.

vector/src/main/java/im/vector/RegistrationManager.java
```diff
@@ -0,0 +1,925 @@
...
+                        }
+
+                        @Override
+                        public void onUnexpectedError(Exception e) {
+                            listener.onThreePidRequested(pid);
+                        }
+
+                        @Override
+                        public void onMatrixError(MatrixError e) {
+                            listener.onThreePidRequested(pid);
```
ylecollen: i assume we should manage the error cases.
We could try again 5 seconds later.

PR #1019, opened 2017-03-14
Login screen: separate field for phone number

sabrinaj:


vector/src/main/java/im/vector/LoginHandler.java
```diff
@@ -135,36 +133,33 @@ public void onMatrixError(MatrixError e) {
...
-     * @param context
-     * @param hsConfig the homeserver config
-     * @param login the login
+     * @param hsConfig
+     * @param username
+     * @param phoneNumber
+     * @param phoneNumberCountry
      * @param password
      * @param callback
      */
```
ylecollen: parameter descriptions ?

PR #1034, opened 2017-03-16
add_unread_badge

ylecollen:
add the new unread messages counter bagde

### DIFF
--- a/vector/src/main/java/im/vector/view/UnreadCounterBadgeView.java
+++ b/vector/src/main/java/im/vector/view/UnreadCounterBadgeView.java
@@ -18,6 +18,7 @@ package im.vector.view;
 import android.content.Context;
 import android.graphics.Bitmap;
 import android.net.Uri;
+import android.support.annotation.IntDef;
 import android.util.AttributeSet;
 import android.view.View;
 import android.widget.ImageView;
@@ -26,11 +27,24 @@ import android.widget.TextView;
 
 import org.w3c.dom.Text;
 
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
 import im.vector.R;
 
 public class UnreadCounterBadgeView extends RelativeLayout {
 
+    // the background settings
+    public static final int HIGHLIGHTED = 0;
+    public static final int NOTIFIED = 1;
+    public static final int DEFAULT = 2;
+
+    @IntDef({HIGHLIGHTED, NOTIFIED, DEFAULT})
+    @Retention(RetentionPolicy.SOURCE)
+    public @interface Status {}
+
     private TextView mCounterTextView;
+    private View mParentView;
 
     public UnreadCounterBadgeView(Context context) {
         super(context);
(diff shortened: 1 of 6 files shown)


vector/src/main/res/layout/unread_counter_badge.xml
```diff
@@ -0,0 +1,17 @@
...
+    android:layout_width="25dp"
+    android:layout_height="12dp">
+
+    <TextView
+        android:id="@+id/unread_counter_badge_text_view"
+        android:layout_centerInParent="true"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:textStyle="bold"
+        android:textSize="7dp"
```
sabrinaj: "sp" unit is recommended.

> Scale-independent Pixels - This is like the dp unit, but it is also scaled by the user's font size preference. It is recommend you use this unit when specifying font sizes, so they will be adjusted for both the screen density and the user's preference.

vector/src/main/res/layout/unread_counter_badge.xml
```diff
@@ -0,0 +1,17 @@
...
+        android:text="127" />
```
sabrinaj: Should be `tools:text="127" `

vector/src/main/res/layout/unread_counter_badge.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout
```
sabrinaj: Maybe the custom view could be simplified as a Textview?

vector/src/main/java/im/vector/view/UnreadCounterBadgeView.java
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * Update the badge value and its status
+     *
+     * @param value the new value
+     * @param status the new status
+     */
+    public void updateCounter(int value, @Status int status) {
+        if (value > 0) {
+            if (value > 999) {
+                mCounterTextView.setText("999+");
```
sabrinaj: Should be displayed with "k". "1.2k" should be displayed for 1200 for example.

vector/src/main/java/im/vector/view/UnreadCounterBadgeView.java
```diff
@@ -0,0 +1,101 @@
...
+                mCounterTextView.setText(value + "");
+            }
+            setVisibility(View.VISIBLE);
+
+            GradientDrawable shape = new GradientDrawable();
+            shape.setShape(GradientDrawable.RECTANGLE);
+            shape.setCornerRadius(100);
+            if (status == HIGHLIGHTED) {
+                shape.setColor(getContext().getResources().getColor(R.color.vector_fuchsia_color));
+            } else if (status == NOTIFIED) {
```
sabrinaj: Deprecated I think. Can use `ContextCompat.getColor(getContext(), R.color.vector_fuchsia_color)` instead

sabrinaj:
Merged in ux_rework

PR #1050, opened 2017-03-21
Fix issue with registration on custom servers

sabrinaj:


vector/src/main/java/im/vector/RegistrationManager.java
```diff
@@ -176,6 +176,18 @@ public void setCaptchaResponse(final String captchaResponse) {
     }
 
     /**
+     * Set the supported flow stages for the current home server)
+     *
```
ylecollen: a ')' without a '('

Issue #1069, opened 2017-03-25
Does not send quick-reply when sending to room with unverified devices


krombel:
When one room member changes/adds a devices so the key was not seen before sending the quick reply (or when I for myself reinstalled the device) and then get a message I am offered to send a quick-reply. But this reply will not be send. Additionally I do not get the hint that the message could not be delivered - it just disappears

Matrixcoffee:
Perhaps the "Quick Reply" feature should just ignore (temporarily blacklist) any new devices, and send the message anyway. It will be more in line with what people expect.

ylecollen:
@Matrixcoffee 
The android client already tries to send the message but there was no feedback that the message sending fails.

The message should be displayed in read (unsent). 
It seems working on my side.

But, we need to add an UI item to warn there are some unsent items in a dedicated room (from the recent page).

PR #1073, opened 2017-03-27
fix_unknown_devices_quick_reply

ylecollen:
Display an alert when some unknown devices are found while using the quick replay
Should fix https://github.com/vector-im/riot-android/issues/1069

### DIFF
--- a/vector/src/main/java/im/vector/activity/LockScreenActivity.java
+++ b/vector/src/main/java/im/vector/activity/LockScreenActivity.java
@@ -26,6 +26,7 @@ import android.os.Bundle;
 import android.text.TextUtils;
 import android.text.TextWatcher;
 
+import org.matrix.androidsdk.crypto.MXCryptoError;
 import org.matrix.androidsdk.util.Log;
 
 import android.view.View;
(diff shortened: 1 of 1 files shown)


krombel:
Whats about the issue that the message is lost in this case?

ylecollen:
@krombel 
The message won't be lost : it will be stored in the dedicated room history as "unsent".
The user will have to validate the unknown devices list before sending them.

krombel:
I could not find the message that was send via quick-reply in the history. It seems that the message somehow was added to local room history but I cannot see it. Just the resend possibility is shown.
After choosing the resend option I could see the message again
(Just tested a few moments ago with current dev and stable)

ylecollen:
@krombel 
When you says "history", do you means the home/recents history or the room history (i.e the room page) ?

krombel:
I mean the list of messages that contains the send/received messages inside a room

ylecollen:
@krombel  
on my side, the message is displayed at the bottom in red.

krombel:
I tested again and the message was not there when I send thequick-reply and checked instantly.
But now I see the unsend message when I left and refocused the room

PR #1082, opened 2017-03-29
Spanish Translation

general50:
Work In Progress!
planned completion before 01/04/2017 (dd/mm/yy)


general50:
Sorry about all the commits. I didn't know it would do that.

general50:
Date for planned completion should read 1/4/2017 (dd/mm/yy)  not 4/1/2017. Pardon my noobness.

PR #1104, opened 2017-04-07
Text with auto complete

ylecollen:
Add the matrix user ID / username auto-completion in text edit in the room activity.


vector/src/main/java/im/vector/adapters/AutoCompletedUserAdapter.java
```diff
@@ -0,0 +1,187 @@
...
+                        if ((null != user.displayname) && user.displayname.toLowerCase().startsWith(prefixString)) {
+                            newValues.add(user);
+                        }
+                    }
+                }
+            }
+
+            // sort the results
+            if (mIsSearchingMatrixId) {
+                Collections.sort(newValues, new Comparator<User>() {
```
sabrinaj: Comparator could be created only once outside the filter class instead of being recreated at each filtering

PR #1108, opened 2017-04-07
Notifications improvements

erikjohnston:
This is a PR that tests out changing the notifications to be more useful. It is currently quite hacky and needs quite a bit of cleanup, but demonstrates what it looks like and how it should (roughly) work.

This currently only looks good on Android N.

Features:
- Collapsible notifications on a per room basis
- Colours the notifications depending on if there was a bing or not
- Orders the notifications depending on if there were bings in them or not.

Incomplete list of things to do:
- Generally do this in a less hacky way
- Handle older androids nicer, currently there's just a placeholder
- Handle 1:1 rooms in a better way
- Add some config options

Screenshots, collapsed view and then expanded view:

collapsed
expanded

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.3'
+        classpath 'com.android.tools.build:gradle:2.3.1'
         classpath 'com.google.gms:google-services:1.3.0'
         // NOTE: Do not place your application dependencies here; they belong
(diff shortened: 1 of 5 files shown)


ylecollen:
For android < N devices, the notification will the following content
to summarise the expected behaviour (android < N)
1- A line per room
2- The noisy ones first (displayed in red)
3- the syntax is <Room name> : (<sender>) the messages if there is only one message or <Room name> : XXX unread messages
4- display the quick reply button if there are unread messages from only one room.

PR #1115, opened 2017-04-11
Escape single quotes in Portuguese translations

dtygel:
Two strings had unescaped single quote inside the translation. Fixed.


ylecollen:
i already fixed it.

PR #1119, opened 2017-04-12
Notification improvements

ylecollen:
Improve the notifications display.

They contain either a status for the notified rooms list or the latest notified messages if there is only one room.


ylecollen:
notification_several_rooms
notifications_one_room

PR #1125, opened 2017-04-14
Ux rework add badges

ylecollen:
Add unread messages counter badges

### DIFF
--- a/vector/src/main/java/im/vector/activity/VectorHomeActivity.java
+++ b/vector/src/main/java/im/vector/activity/VectorHomeActivity.java
@@ -26,9 +26,10 @@ import android.content.IntentFilter;
 import android.content.SharedPreferences;
 import android.net.Uri;
 import android.os.Bundle;
-import android.os.Handler;
 import android.preference.PreferenceManager;
 import android.support.annotation.NonNull;
+import android.support.design.internal.BottomNavigationItemView;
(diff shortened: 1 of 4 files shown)


vector/src/main/java/im/vector/activity/VectorHomeActivity.java
```diff
@@ -1442,6 +1465,202 @@ public void run() {
...
+            if ((null == filteredRoomIds) || !filteredRoomIds.isEmpty()) {
+                for(RoomSummary summary : summaries) {
+                    Room childRoom = mSession.getDataHandler().getStore().getRoom(summary.getRoomId());
+
+                    // test if the room is allowed
+                    if ((null != childRoom) && ((null == filteredRoomIds) || filteredRoomIds.contains(summary.getRoomId()))) {
+                        highlightCount += childRoom.getHighlightCount();
+                        notificationCount += childRoom.getNotificationCount();
+                        mustBeHighlighted |= summary.isHighlighted();
+                    }
```
sabrinaj: Maybe each fragment could provide itself the list of rooms it handles (filteredRoomIds)

ylecollen: The fragments are not always created whereas we need to manage the tabbar.

PR #1159, opened 2017-04-25
Improve bug report

ylecollen:
Implement the new bug report format.

### DIFF
--- a/vector/src/main/java/im/vector/util/BugReporter.java
+++ b/vector/src/main/java/im/vector/util/BugReporter.java
@@ -18,29 +18,16 @@
 package im.vector.util;
 
 import java.io.BufferedReader;
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.DataOutputStream;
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/util/BugReporter.java
```diff
@@ -622,8 +511,44 @@ private static Bitmap takeScreenshot() {
...
+
+    /**
+     * Save the logcat
+     *
+     * @param context       the context
+     * @param isErrorLogcat true to save the error logcat
+     * @return the file if the operation succeeds
+     */
+    private static File saveLogCat(Context context, boolean isErrorLogcat) {
+        File logCatErrFile = new File(context.getCacheDir().getAbsolutePath(), isErrorLogcat ? "logCatError.log" : "logError.log");
```
sabrinaj: constants for file names ?

Matrixcoffee:
FYI: this PR adds compression to the bug reports, so it sends faster and consumes less network data. The reduction is a whopping factor 5 according to @ylecollen.

PR #1164, opened 2017-04-26
ux_rework_improve_public_rooms

ylecollen:
Remove the directory source search and replace by a + button

### DIFF
--- a/vector/src/main/java/im/vector/activity/RoomDirectoryPickerActivity.java
+++ b/vector/src/main/java/im/vector/activity/RoomDirectoryPickerActivity.java
@@ -16,22 +16,22 @@
 
 package im.vector.activity;
 
-import android.app.SearchManager;
 import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
-import android.support.v4.view.MenuItemCompat;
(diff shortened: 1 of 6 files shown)


vector/src/main/java/im/vector/activity/RoomDirectoryPickerActivity.java
```diff
@@ -210,6 +199,64 @@ public void onUnexpectedError(Exception e) {
...
+                        public void onSuccess(Integer count) {
+                            Intent intent = new Intent();
+                            intent.putExtra(EXTRA_OUT_ROOM_DIRECTORY_DATA, new RoomDirectoryData(serverUrl, serverUrl, null, null, false));
+                            setResult(RESULT_OK, intent);
+                            finish();
+                        }
+
+                        private void onError(String error) {
+                            Log.e(LOG_TAG, "## onSelectDirectoryServer() failed " + error);
+                            mLoadingView.setVisibility(View.GONE);
```
sabrinaj: Maybe activity has been killed before we get the request response, which would cause a crash

ylecollen: it does not seem to crash.

PR #1202, opened 2017-05-05
Add menu option to check for updates for jenkin builds

erikjohnston:
Setting `JOB_URL` and `BUILD_NUMBER` environment variables during build will enable a menu item that will check for any updates and prompt if the user wants to download them.

These enviroment variables are enabled by default for jenkins builds.

---

This is a quick hacky first attempt at doing this, and I'm not sure whether this is even something we want to do but I think it might be quite useful :)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -22,6 +22,10 @@ org.gradle.daemon=true
 org.gradle.parallel=true
 org.gradle.configureondemand=true
 
+# Can manually override the config options for "check for updates function"
+# JENKINS_JOB_URL = "https://matrix.org/jenkins/job/VectorAndroidDevelop/"
+# JENKINS_BUILD_NUMBER = 1246
+
 # To do a signed release APK, put these variables in ~/.gradle/gradle.properties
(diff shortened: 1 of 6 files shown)


erikjohnston:
(This probably should have some comments...)

ylecollen:
1- Don't you think if you would be better to have a beta on google play as we talked about ?

2 - Do you think it would be possible to have a notification when there is a new jenkins build ?
This notification would be sent everyday (if there is an update)

erikjohnston:
> 1- Don't you think if you would be better to have a beta on google play as we talked about ?

It looks like it will lag a couple of days behind, so will be annoying if you want to update immediately. :(

> Do you think it would be possible to have a push when there is a new jenkins build ?
This push would be sent everyday (if there is an update).

I was thinking about that, but figured itd still need the same code (ignoring ui) so could be added later.

We csn certainly configure to only display on debug builds by removing the jenkins config from the release config. I was trying to put most of the stuff into one class so as to not complicate the rest of the code, but that doesnt work for ui.

Issue #507, opened 2016-09-15
Dark and compact theme on mobile


AmandineLP:
Compact = without profile picture

rrthomas:
A dark theme would be great!

ptman:
I'd like a compact theme with less whitespace to fit more text on the screen on mobile

ara4n:
increasing the priority because a) this wastes battery on OLED devices, b) all the upvotes, c) it looks pretty

jgkamat:
I'm working on a theme switcher right now, hopefully I'll have something basic (without colors for now at least) in a PR later today or tomorrow! :smile:

ara4n:
We have an official design for the dark theme - the artwork looks like this (and should match the dark theme on Riot/Web):

<img width="371" alt="screen shot 2017-05-28 at 23 39 38" src="https://cloud.githubusercontent.com/assets/1294269/26532663/ed4b7f30-43fe-11e7-80d4-afd041d38d80.png">

jgkamat:
I think this can be closed now :tada:

ptman:
There is a font size setting, but the spacing is still a bit "modern", i.e. anything but compact

moshpirit:
Also in settings the huge space between sections is still white instead of black. 
![](https://i.imgur.com/pG3vdIE.png)

okmrponcho:
How about night mode for ios?

PR #1240, opened 2017-05-29
Add a basic theme switcher

jgkamat:
Please don't merge this yet, it's still a WIP!

This PR adds a simple theme switcher to the option menu, and changes a single color (the `room_background` color is changed from white to black) to demonstrate that it works. I don't want to change more than a single color right now since that would require a lot of changes (basically, moving where colors are defined into the themes file, rather than individual components or `color.xml`.)

(In theory) the light theme should be the current default theme, while the dark theme is the light theme with a 'dark android base' (so that unstyled elements use a dark theme), and making the room background pure black.

Once this gets a :+1: from you guys (revewiing just the theme switcher itself and the structure of how the colors are going to be laid out), I'll go ahead and to implement the dark theme that is pictured below.

This is my first time ever working on an android app, so let me know if I'm doing anything that looks wrong! I'm going to try and leave a few comments on here to point out things I'm not sure about.

Inspiration and guidance were taken from both #48 and [k9 mail](https://github.com/k9mail/k-9), but I ended up writing everything that's here by myself. I'm pretty busy over the next couple of weeks, so I might be slow to respond, but I will (hopefully) eventually push this through! :smile: 

See #48, #507 

Here are some screenshots:
Light theme (current default):
2017-05-28-184559_422x814_scrot

Dark Theme (just changing the background to black, it's going to get better :P)
2017-05-28-184636_417x825_scrot

Target (from @ara4n):
 target

(also, I'm @jgkamat:matrix.org if you want to ping me :D)


vector/src/main/java/im/vector/activity/CountryPickerActivity.java
```diff
@@ -72,6 +73,7 @@ public static Intent getIntent(final Context context, final boolean withIndicato
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        ThemeUtils.activitySetTheme(this);
```
jgkamat: I can remove all of these repeated `activitySetTheme` calls if I make a common superclass to all Vector activities, does that sound like a good idea to you?

vector/src/main/res/values/themes.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <!--TODO decide whether these themes are the right base-->
+    <style name="Theme.Vector.Light.Base" parent="@style/Theme.AppCompat.Light.NoActionBar" />
+    <style name="Theme.Vector.Dark.Base" parent="@style/Theme.AppCompat.NoActionBar"/>
+
+    <style name="Theme.Vector.Light.Common" parent="Theme.Vector.Light.Base">
```
jgkamat: I took this idea of redirecting colors from k9 mail. This would require me to change every color reference into an attribute reference here. Does anyone have a better way to do this which would require changing less code? If not, it won't be too bad on my end (I can write a couple of scripts), but this would probably cause a bunch of merge conflicts...

vector/src/main/java/im/vector/VectorApp.java
```diff
@@ -393,6 +394,15 @@ private void stopActivityTransitionTimer() {
 
         mIsCallingInBackground = false;
         mIsInBackground = false;
+        setTheme();
+    }
+
+    private void setTheme() {
```
jgkamat: I'm not sure if this can be placed anywhere better than here. Any ideas?

jgkamat:
Also, one more question: should I rebase this onto the ux rework branch? If that would help (or if it sounds like a good idea) let me know. =)

jgkamat:
I've just rebased this onto ux_rework, and it seems to still work the same, so let me know if I'm doing anything wrong! If you think this looks good, I'll begin working on adding the colors.

.editorconfig
```diff
@@ -0,0 +1,19 @@
+root = false
+
```
sabrinaj: That file should not be commited.

vector/src/main/java/im/vector/VectorApp.java
sabrinaj: I think you could do that in your `ThemeUtils` class. I.e, maybe the first time you try to apply the theme to an activity, if `currentTheme`  is null, you could then retrieve the value in SharedPreference (or give it a default value).

sabrinaj: I'd also suggest using a specific key (constant) to put in SharedPref instead of using a string that might be changed or deleted.

vector/src/main/java/im/vector/activity/CountryPickerActivity.java
sabrinaj: It might be a good idea, yes.

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.java
```diff
@@ -290,6 +292,25 @@ public boolean onPreferenceClick(Preference preference) {
             });
         }
 
+        // Themes
+        ListPreference themePreference = (ListPreference) findPreference("Theme");
```
sabrinaj: Use a string resource or a constant instead of a hard coded value.

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.java
```diff
@@ -290,6 +292,25 @@ public boolean onPreferenceClick(Preference preference) {
...
+        if (null != themePreference) {
+            themePreference.setOnPreferenceChangeListener(new Preference.OnPreferenceChangeListener() {
+                @Override
+                public boolean onPreferenceChange(Preference preference, Object newValue) {
+                    if (newValue instanceof String) {
+                        ThemeUtils.setTheme((String)newValue);
+                    }
+                    // Display a message to the user
+                    CharSequence text = "You may need to restart the app to see these changes.";
```
sabrinaj: Move the text in `string.xml`

vector/src/main/java/im/vector/util/ThemeUtils.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2016 OpenMarket Ltd
```
sabrinaj: ` * Copyright 2017 Vector Creations Ltd` instead

vector/src/main/java/im/vector/util/ThemeUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private static Integer currentTheme = null;
+
+    // Maps theme names to formal ID's
+    private static HashMap<String, Integer> themeMap = new HashMap<>();
+    static {
+        themeMap.put("dark", R.style.Theme_Vector_Dark);
+        themeMap.put("light", R.style.Theme_Vector_Light);
+    }
+
```
sabrinaj: Maybe you could use `IntDef` instead? 
https://developer.android.com/reference/android/support/annotation/IntDef.html

vector/src/main/res/values/themes.xml
sabrinaj: I can't think of a better way right now. But I don't think so many colors have to be changed?

jgkamat: Unfortunately, I think that all the colors will need to be changed (since the dark theme 'inverts' all the colors). I'd like to make all the colors attributes anyway, so they can be changed easily in the future. 

I'm making a [script to help me do this though](https://gist.github.com/jgkamat/59f041d7cfe45cb2b0b757227680b9df), so I can keep re-applying the changes while avoiding merge conflicts.

jgkamat:
Thanks for the review @sabrinaj, these are all good points (I'm very new at developing android, sorry). I'll go ahead and fix all of these problems and begin adding the actual colors! :smile:

jgkamat:
For some reason, which I can't find out, applying any theme over the `VectorBaseSearchActivity` causes [this line](https://github.com/jgkamat/riot-android/blob/master/vector/src/main/java/im/vector/activity/VectorBaseSearchActivity.java#L69) to return null, which causes a crash. Do you know why `getSupportActionBar()` would return null when I apply a theme over it? Once I figure that out, I think I'll be able to convert my references script and begin applying a real theme.

sabrinaj:
The parent of your style `Theme.Vector.Light` is `Theme.AppCompat.Light.NoActionBar` so it makes sense that you get null when you try to get the action bar. Currently, in the activity `VectorUnifiedSearchActivity`, we apply the `SearchesAppTheme` style which inherits from `Theme.AppCompat.Light`.

jgkamat:
Hmm, that poses an interesting problem then.

I would like to only have two base themes (dark and light), but if some activities require the action bar, and the action bar must be set in the theme, we would have to make two themes (a dark actionbar, dark noactionbar, light actionbar, light noactionbar). As far as I know, there's no way to make android apply two styles.

Is it possible to override the current style for one property (ie: the action bar), so I can programatically re-add the action bar after applying the no-actionbar style? That would let me keep my two base themes, and avoid creating duplicates. 

I'm very new to this, so I might be overlooking something much easier though, so let me know if that's the case! I'm going to ask around and see if anyone else I know has any ideas to do this cleanly.

jgkamat:
I'm really close to this being finished. I'm still a little bit blocked on the AppCompat themes (since most of the activities need NoAppCompat, I'm sticking to that as the base). For the room details, I'm simply turning off all of the theming (because I can't figure out how to convert that to a TabLayout). For the search menu, I tried adding a new toolbar manually, which works fine, but the textedit seems to no longer be responding to input (it's not actually searching anymore). If you could help me on these two issues I'd be really happy!

Not all the screens have the dark theme enabled, but this is starting to get quite big, so I want to just get the basics out, and I can probably follow in future PR's.

To generate the last commit, please run this [conversion script](https://gist.github.com/jgkamat/59f041d7cfe45cb2b0b757227680b9df) to convert the hard coded links to theme links. I made a script because I anticipate that commit causing a bunch of merge conflicts. 

The results so far:

2017-07-08-181934_394x771_scrot
2017-07-08-181941_375x765_scrot
2017-07-08-182037_419x777_scrot
2017-07-08-182105_397x766_scrot

I would like to get this merged as soon as possible, to avoid conflicts, so please let me know if you have any feedback, or any idea on how to fix these issues.

sabrinaj:
As a reminder, here is a design that was made for the dark theme:
image
The UI evolved since but it gives an idea of the expected colors.

For the toolbar, for example, I'm not sure we want that new green color since the current one is part of Riot main colors. So I'd suggest that you use the color from the design above or, if theming it causes some issues, keep the toolbar as it is in light theme.

vector/src/main/java/im/vector/activity/VectorBaseSearchActivity.java
```diff
@@ -66,11 +67,19 @@
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
+        setContentView(R.layout.activity_vector_unified_search);
```
sabrinaj: This layout has nothing to do here. This activity should probably be declared abstract since it cannot be used directly. Calling setContentView() knowing that it'll be overriden by the activity extending this class doesn't feel right and is expensive.

sabrinaj:
I'm not sure it'll be possible to keep only two base themes for all activities. For activities having there own theme, a dark version of it will be necessary. Maybe an option would be, in VectorActivity, to add an abstract method that every activity would define to specify its own light/dark theme?

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -2214,7 +2224,8 @@ public void onClick(View v) {
 
                 if (unreadCount > 0) {
                     iconId = R.drawable.newmessages;
-                    textColor = R.color.vector_fuchsia_color;
+                    textColor = ThemeUtils.getColor(this, R.attr.vector_fuchsia_color);
```
sabrinaj: Some renaming could be done on attributes. Fuchsia is a color so its value shouldn't change depending on the theme. Ideally, the attribute should describe what it is, like "room background".
Many things are not very well named currently, that theme update could be an opportunity to rename some. :-)

sabrinaj:
On my device with Android 6, some parts (status bar, lock icon, memberships messages) are in black:
image

vector/src/main/java/im/vector/activity/VectorBaseSearchActivity.java
jgkamat: Ah, sorry! I didn't even notice that this was being set in the superclass (and I got really confused). I'll fix this in the next update :smile:

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
jgkamat: While I definetly agree with you that renaming these is a good idea, I would like to try and keep this PR as small as possible (to get something basic out without having tons and tons of merge conflits), so I would prefer to keep this as is for now. I would be happy to submit more follow up PR's trying to refactor the differences out of things like `vector_fuchsia_color` though. 

If you think I should go ahead anyway, let me know though! :smile:

jgkamat:
@sabrinaj, hmm, I think that given the state of things currently (how every activity has it's own theme), I think that that the dark/light theme method in the abstract class seems like a good way to solve the problem. I was hoping to avoid something like this, because it would make adding future themes very difficult, but I'd rather keep the UI as similar to it is currently, so I'll probably re-work the last few commits to use that method instead.


RE: the black text on join/part, It's very possible I missed a few colors, so thanks for letting me know! I'll check in on that the next time I have a chance to work on this and see if I can get that fixed!

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
sabrinaj: Sure, it can be done later on.

sabrinaj:
@jgkamat To add new themes in the future, a possibility may be that activities having their own theme would hold a map of it. That abstract method in VectorActivity could then have a parameter to specify which theme we are looking for.

Something like:
```
public class VectorUnifiedSearchActivity ... {
    private static final Map<Integer, Integer> themeMap= new HashMap<>();
    static {
        themeMap.put(ThemeUtils.THEME_LIGHT, R.style.SearchesAppTheme);
        themeMap.put(ThemeUtils.THEME_DARK, R.style.SearchesAppTheme_Dark);
    }

    @Override // Abstract method from VectorActivity
    public int getTheme(@ThemeUtils.ThemeList final int themeKey){
        return themeMap.get(themeKey);
    }
```
(Just an idea)

jgkamat:
@sabrinaj, that sounds like a good idea, that's probably the way I'll go. It'll also allow themes to 'opt out' of implementing some themes if they want to (just return light, for example). 

As for avoiding styling the status bar, the status bar looks really bad on the dark theme (almost unreadable) if it's not themed at all, I'd really like to theme it. I'll try to figure out what the issues are behind the toolbar light theme (I think it's mainly with the search text color), and see if I can keep it the same as before.

jgkamat:
I've taken the easy way out for now and just left the old styling on activites that needed it. The 'override method' way of doing it didn't seem to work, since the proper 'themed' elements needed to be present for those (so what I'm doing is just calling setTheme on any activities that must have an action bar). 

I also tried theming the home actionbar to look identical to the current version in light mode, let me know if you find any more issues with that. I don't think that I can turn off theming just for the action bar, the parent theme seems to be overriding it.

I don't know how to fix the 'black images' issue, the images seem to be black in source, and I'm not sure how (is it possible?) to recolor them to be white.

I'm not getting the join/parts being black anmore, can you try that again and see if it's themed properly for you? 

Lastly, I would like to keep the current theme colors (darker than the mock up), because I would like to have the same color scheme as riot web, for consistency. Most of the colors were pulled from riot web. I feel like the mock up's background is too light (it's a little hard to read text on it for me). If you think the mock up is the way to go though, I can try to tweak the colors to be more like that. 

One last point, I'll be offline two week's from now, so if this dosen't get merged by the end of this week, I'll come back online about 3-4 weeks from now, and hopefully I'll be able to finish the job! If anyone wants to clean this up or tweak it while I'm offline, that would be amazing too! :smile: 

Let me know what you think! :tada:

sabrinaj:
It seems that the mockup is incorrect so following riot-web about the colors is the right thing to do, my bad.

For the black images, I think the best option would be to use the [tint](https://developer.android.com/reference/android/widget/ImageView.html#attr_android:tint) attribute, maybe in a class extending `ImageView`.

jgkamat:
I just pushed a few commits trying to tint the encryption icons, and I think it should be good on both dark and light themes. 

I also noticed that the black text that you pointed about above happens when you switch themes without a full restart. If you do a full restart, the text becomes normal. I think that's related to setting the message text programatically, which I might take a look at later.

sabrinaj:
We cannot merge something on develop that is not fully done and tested so I created a new branch `dark_theme` and merged your pull request on it. If you are willing to continue, you can then create a new pull request on the dark theme branch. 
Note that this feature became one of the priorities so there is a possibility that Yannick works on it during your absence.

Also, could you add the _Sign off_ comment as specified in the file below?  
https://github.com/vector-im/riot-android/blob/master/CONTRIBUTING.rst

Thank you for your contribution on that feature. :-)

jgkamat:
Not sure if this is all I need to do but:

```
Signed-off-by: Jay Kamat <jaygkamat@gmail.com>
```

Let me know if there's anything else I need to do to sign off. 

If you need any extra help just ping me, comment here, or create an issue  and assign me, and I'll try to submit more PR's to the dark_theme branch! Thanks for working on this, it means a lot to me 🎉 (and I really want to see  #507 closed!)

sabrinaj:
It's all you needed to do, yes, thank you. 

I just added you to in authors on the dark_theme branch.
https://github.com/vector-im/riot-android/blob/dark_theme/AUTHORS.rst

lgleasain:
I'll pick things up for a bit.

PR #1256, opened 2017-05-30
Update ES strings translation and fix typo in ES_rMX strings

general50:


general50:
Incomplete.

general50:
Ok, looks good now.

PR #1287, opened 2017-06-06
Unread msg management

ylecollen:


ylecollen:
one case the read marker should not be displayed
-> scroll back in the room history
-> receive one new event
-> click on the red back arrow to display the latest messages

A read marker is displayed after the latest event

ylecollen:
Another case
Open the room activity
put the app in background
send > 10 messages in the room

The read marker is not displayed at the right position

ylecollen:
Another case
Scroll to back in the history
Receive 5 new messages 
scroll down

The read marker is only displayed when the history reaches the latest event

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -927,9 +968,48 @@ public void onDestroy() {
...
+                mRoom.setReadMakerEventId(newReadMarkerEventId);
+            }
+        }
+    }
+
     @Override
     protected void onPause() {
         super.onPause();
+
+        if (mVectorMessageListFragment.getMessageAdapter().isUnreadViewMode()) {
```
sabrinaj: Done

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -1202,6 +1283,118 @@ public void onScroll(int firstVisibleItem, int visibleItemCount, int totalItemCo
...
+        boolean hasUnreadMessageHidden = false;
+        RoomSummary summary = mRoom.getDataHandler().getStore().getSummary(mRoom.getRoomId());
+        if (summary != null) {
+            String readMarkerEventId = summary.getReadMarkerEventId();
+            String readReceiptEventId = summary.getReadReceiptEventId();
+            Log.d(LOG_TAG, "## checkUnreadMessage readMarkerEventId " + readMarkerEventId+" readReceiptEventId "+readReceiptEventId);
+
+            synchronized (this){
+                if (readMarkerEventId != null && !readMarkerEventId.equals(readReceiptEventId)) {
+                    if (readMarkerEventId.startsWith(mRoom.getRoomId())) {
```
sabrinaj: Done

PR #1293, opened 2017-06-07
ux_rework_no_local_contacts

ylecollen:
Replace "no local contacts" when the contacts permission is not granted.


vector/src/main/java/im/vector/adapters/PeopleAdapter.java
```diff
@@ -57,6 +58,8 @@
 
     private final OnSelectItemListener mListener;
 
+    private Context mContext;
```
sabrinaj: I think we should avoid as much as possible holding a reference on `Context` since it can easily lead to memory leaks. I'd suggest adding the context as a parameter of `setLocalContacts` instead. Or maybe use 
  `@BindString(R.string.title) String title;` to have the two placeholders accessible at all time without having to do a `getString(...)`.

Issue #1276, opened 2017-06-06
Opened app and it started resending a picture I had already sent to a room


ylecollen:
Reported by @vurpo:vurpo.fi
https://github.com/matrix-org/riot-android-rageshakes/issues/235

vurpo:
This has happened a couple of times before, but I don't know how to reproduce it. Usually when it happens, I've shared a picture to a room, then exited the app (home or switch to another app), then later opened the app again using the home screen icon or the app switcher. Sometimes instead of sending the picture again, it instead displays the "share to room" list asking me which room to share to.

ylecollen:
Richvdh got the same issue but i don't see how it is possible.

ylecollen:
Steps to reproduce
1- share a file with Riot
2- open Riot : the rooms chooser is opened. Select a room and go back to the home screen
3- put the application in background
4- kill the application with "Devices monitor"
5 - Resume the application

The ActivityManager saves the HomeActivity state (ie the intent) when it is recreated (because it has been killed).

There is a "share file" extra params which is retrieved and the files chooser is displayed again.
 

https://stackoverflow.com/questions/16837595/will-bundle-savedinstancestate-be-alive-after-application-is-being-killed

ylecollen:
fixed on develop

PR #1311, opened 2017-06-12
fix_auto_resend_files

ylecollen:
fix the issue #1276 Opened app and it started resending a picture I had already sent to a room.

### DIFF
--- a/vector/src/main/java/im/vector/activity/VectorHomeActivity.java
+++ b/vector/src/main/java/im/vector/activity/VectorHomeActivity.java
@@ -89,6 +89,7 @@ import im.vector.view.VectorPendingCallView;
  * new rooms.
  */
 public class VectorHomeActivity extends AppCompatActivity implements VectorRecentsListFragment.IVectorRecentsScrollEventListener {
+
     private static final String LOG_TAG = "VectorHomeActivity";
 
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/activity/VectorHomeActivity.java
```diff
@@ -287,32 +285,55 @@ public void onClick(DialogInterface dialog, int which) {
...
         }
         intent.removeExtra(EXTRA_WAITING_VIEW_STATUS);
 
-        mAutomaticallyOpenedRoomParams = (Map<String, Object>) intent.getSerializableExtra(EXTRA_JUMP_TO_ROOM_PARAMS);
-        intent.removeExtra(EXTRA_JUMP_TO_ROOM_PARAMS);
+        // fix issue #1276
+        // if there is a saved instance, it means that onSaveInstanceState has been called.
+        // theses parameters must only be used at activity creation.
+        // The activity might have been created after being killed by android while the application is in background
+        if (null == savedInstanceState) {
```
sabrinaj: Maybe some refactoring could be done to avoid so many  `if (null == savedInstanceState)`. It would make it easier to maintain.

Issue #1330, opened 2017-06-15
Using the name completion as the first item of the message should add a colon (:)


Matrixcoffee:
It is especially annoying because a backspace erases the whole name. I have to use cursor movement to add the : which is difficult on Android.

PR #1349, opened 2017-06-19
add_colon_on_first_item

ylecollen:
fix https://github.com/vector-im/riot-android/issues/1330
Using the name completion as the first item of the message should add a colon (:)


vector/src/main/java/im/vector/view/VectorAutoCompleteTextView.java
```diff
@@ -189,6 +192,36 @@ private void adjustPopupSize() {
...
+    @Override
+    protected void replaceText(CharSequence text) {
+        String before = getText().toString();
+        super.replaceText(text);
+
+        if (mAddColonOnFirstItem) {
+            try {
+                Editable editableAfter = getText();
+
+                // check if the inserted becomes was the new first item
```
sabrinaj: becomes was ?

Issue #1358, opened 2017-06-21
the (+) in the home page covers some rooms 


ylecollen:

ylecollen:
![plus button](https://user-images.githubusercontent.com/8971031/27425688-8cfa56f8-5739-11e7-9c68-8d3cbbf28ea1.jpeg)

Matrixcoffee:
Perhaps the (+) should only be shown if we are scrolled up to the top position.

That might still cause problems if the screen is so large or the list is so small that it fits entirely on the screen, though.

Maybe it should also not be shown if the bottom category is scrolled sideways. It could still fail but it should prevent the issue in most cases.

ylecollen:
Sabrina has implemented a trick to fix this issue.
It should be pushed today.

ylecollen:
Fixed on develop

PR #1366, opened 2017-06-23
Home fab visibility

sabrinaj:
Home fragment: hide the floating button when user scrolls on the nested scrollview or does a "fling" gesture.

Solve https://github.com/vector-im/riot-android/issues/1358


ylecollen:
LGTM

PR #1367, opened 2017-06-23
Back to live btn in preview

sabrinaj:
- On room preview (open by "jump to first unread"), if the user hits the "jump to bottom" icon, it now closes the preview to bring back the user **at the bottom** of the live (differs from the back button since the back button closes the preview with the live remaining where it was before opening the preview)

- Add some padding around the "jump to bottom" icon so it's easier to tap on it


ylecollen:
LGTM

PR #1403, opened 2017-07-03
fix https://github.com/vector-im/riot-android/issues/1401

ylecollen:
'clicking on the (+) next to the display of numbers should dump the user in the people section of the settings" "not in the settings section"!

### DIFF
--- a/vector/src/main/res/layout/vector_room_header.xml
+++ b/vector/src/main/res/layout/vector_room_header.xml
@@ -56,56 +56,66 @@
 
     <RelativeLayout
         android:id="@+id/action_bar_header_room_members_layout"
-        android:layout_below="@id/action_bar_header_room_topic"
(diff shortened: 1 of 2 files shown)


vector/src/main/res/layout/vector_room_header.xml
```diff
@@ -56,56 +56,66 @@
...
+        android:layout_below="@id/action_bar_header_room_topic"
+        android:layout_marginBottom="19dp">
 
         <!-- dummy text view to compute the text height -->
         <TextView
             android:id="@+id/action_bar_header_line_height"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:visibility="invisible"
+            android:text="measurement text"
```
sabrinaj: That hardcoded string should be removed. Not sure that view is still useful also.

PR #1429, opened 2017-07-12
Refactor message adapter

ylecollen:
Move the messages adapter from the SDK to the app


vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
```diff
@@ -153,272 +185,456 @@ public VectorMessagesAdapter(MXSession session, Context context, MXMediasCache m
...
+        return ContextCompat.getColor(mContext, R.color.vector_green_color);
     }
 
-    /**
-     * Provides the formatted timestamp to display.
-     * null means that the timestamp text must be hidden.
-     * @param event the event.
-     * @return  the formatted timestamp to display.
+    private int getSendingMessageTextColor() {
+        return ContextCompat.getColor(mContext, R.color.message_sending);
```
sabrinaj: Maybe you could use @BindColor from Butterknife?

vector/src/main/res/layout/message_header.xml
```diff
@@ -0,0 +1,26 @@
...
+
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:id="@+id/messagesAdapter_message_header"
+    android:orientation="vertical">
+
+    <TextView
+        android:id="@+id/messagesAdapter_message_header_text"
+        android:layout_width="fill_parent"
```
sabrinaj: `match_parent` instead

vector/src/main/res/layout/message_header.xml
```diff
@@ -0,0 +1,26 @@
...
+        android:layout_height="wrap_content"
+        android:textAppearance="?android:attr/textAppearanceSmall"
+        android:text = "A message here"
```
sabrinaj: I think we should removed all hard coded strings and use `tools:text`

PR #1438, opened 2017-07-17
Add mels

ylecollen:
First implementation of the membership merge support


vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
```diff
@@ -171,17 +176,20 @@
     private final VectorMessagesAdapterMediasHelper mMediasHelper;
     protected final VectorMessagesAdapterHelper mHelper;
 
+    private final HashSet<String> mHiddenEventIds = new HashSet<>();
```
sabrinaj: `private final Set<String>`?

vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
```diff
@@ -1969,4 +2160,163 @@ public boolean onMenuItemClick(final MenuItem item) {
...
+     */
+    private void removeFromEventGroup(MessageRow row) {
+        if (supportMessageRowMerge(row)) {
+            String eventId = row.getEvent().eventId;
+            for (EventGroup eventGroup : mEventGroups) {
+                if (eventGroup.contains(eventId)) {
+                    eventGroup.removeByEventId(eventId);
+
+                    if (eventGroup.isEmpty()) {
+                        mEventGroups.remove(eventGroup);
```
sabrinaj: Removing the object from the list while iterating on it can cause issues. I think an iterator would be better.
https://stackoverflow.com/questions/1196586/calling-remove-in-foreach-loop-in-java

vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
```diff
@@ -190,28 +190,67 @@ static TextView setTimestampValue(View convertView, String value) {
...
+            roomMember = roomState.getMember(event.getSender());
+        }
+
+        String url = null;
+        String displayName = null;
+
+        // Check whether this avatar url is updated by the current event (This happens in case of new joined member)
+        JsonObject msgContent = event.getContentAsJsonObject();
+
+        if (msgContent.has("avatar_url")) {
```
sabrinaj: Could be a constant

vector/src/main/java/im/vector/util/EventGroup.java
```diff
@@ -0,0 +1,280 @@
...
+ */
+public class EventGroup extends Event {
+
+    // events rows map
+    private final Map<String, MessageRow> mRowsMap;
+
+    // rows list (ordered)
+    private final List<MessageRow> mRows;
+
+    // true the merged events are
```
sabrinaj: To be or not to be :)

"are expanded"?

vector/src/main/java/im/vector/util/EventGroup.java
```diff
@@ -0,0 +1,280 @@
...
+ * A EventGroup is a special event that can contain MessageRows
+ */
+public class EventGroup extends Event {
+
+    // events rows map
+    private final Map<String, MessageRow> mRowsMap;
+
+    // rows list (ordered)
+    private final List<MessageRow> mRows;
+
```
sabrinaj: To avoid storing twice the MessageRow, maybe a LinkedHashMap could be used?

vector/src/main/java/im/vector/util/EventGroup.java
```diff
@@ -0,0 +1,280 @@
...
+    // true the merged events are
+    private boolean mIsExpanded;
+
+    // hidden event ids list
+    private final Set<String> mHiddenEventIds;
+
+    private final Context mContext;
```
sabrinaj: Since it is only used in one method, it might be better to avoid having a reference on it and instead pass it as a parameter of the method.

vector/src/main/java/im/vector/util/EventGroup.java
```diff
@@ -0,0 +1,280 @@
...
+            }
+            onRowAdded(row);
+        }
+    }
+
+    /**
+     * @return true if there is no more items
+     */
+    public boolean isEmpty() {
+        return 0 == mRows.size();
```
sabrinaj: `mRows.isEmpty()`?

vector/src/main/res/layout/adapter_item_vector_message_merge.xml
```diff
@@ -0,0 +1,74 @@
...
+        android:layout_marginRight="16dp"
+        android:orientation="horizontal">
+
+        <View
+            android:id="@+id/messagesAdapter_highlight_message_marker"
+            android:layout_width="6dp"
+            android:layout_height="match_parent"
+            android:layout_marginLeft="5dp"
+            android:layout_marginRight="5dp"
+            android:background="#f00" />
```
sabrinaj: If it is set programmatically anyway, I think it can be removed here? (one less thing to manage for dark theme)

vector/src/main/res/layout/adapter_item_vector_search_message_notice.xml
```diff
@@ -0,0 +1,57 @@
...
+
+    </RelativeLayout>
+
+    <View
+        android:id="@+id/messagesAdapter_top_margin_when_no_room_name"
+        android:layout_width="match_parent"
+        android:layout_height="16dp"/>
+
+    <!-- body -->
+    <RelativeLayout
```
sabrinaj: Not sure we need that RelativeLayout?

vector/src/main/res/layout/adapter_item_vector_search_message_notice.xml
```diff
@@ -0,0 +1,57 @@
...
+        <include layout="@layout/adapter_item_vector_message_notice"
+            android:id="@+id/messagesAdapter_search_body"/>
+    </RelativeLayout>
+
+    <RelativeLayout
+        android:layout_width="match_parent"
+        android:layout_height="18dp">
+        <View
+            android:layout_width="match_parent"
+            android:layout_height="1dp"
```
sabrinaj: Maybe marginTop + marginBottom could be added to the View to remove one level in the hierarchy

vector/src/main/res/layout/vector_message_merge_avatar_list.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
sabrinaj: I think <merge> could be used

vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
ylecollen: it works because of the return.

vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
ylecollen: done

vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
ylecollen: done

vector/src/main/java/im/vector/util/EventGroup.java
ylecollen: done

vector/src/main/java/im/vector/util/EventGroup.java
ylecollen: because there is an addToFront method (it is not possible to insert at a dedicated position in a linkedHashMap)

vector/src/main/java/im/vector/util/EventGroup.java
ylecollen: done

vector/src/main/res/layout/adapter_item_vector_message_merge.xml
ylecollen: done

PR #1448, opened 2017-07-20
Add privacy and UX related options

rkfg:
Allow to disable showing read receipts and showing/sending type notifications, disable markdown with a visual setting (currently is only possible via undocumented /markdown off command), change the message font size.

### DIFF
--- a/vector/src/main/java/im/vector/activity/VectorRoomActivity.java
+++ b/vector/src/main/java/im/vector/activity/VectorRoomActivity.java
@@ -1748,7 +1748,8 @@ public class VectorRoomActivity extends MXCActionBarActivity implements MatrixMe
      */
     private void handleTypingNotification(boolean isTyping) {
         int notificationTimeoutMS = -1;
-        if (isTyping) {
+        boolean typingNotificationsEnabled = PreferenceManager.getDefaultSharedPreferences(getApplicationContext()).getBoolean(getString(R.string.settings_send_typing_notifications), true);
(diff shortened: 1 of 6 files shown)


ylecollen:
i'm implementing the languages selection dialog after i will merge your PR (i will need to modify few things)

rkfg:
I've fixed the preferences keys being locale-dependent and added a couple of new settings for media. Now it only requires 3 taps (if you set the preferred defaults) to take and post a photo instead of 5. Much more convenient.

rkfg:
Oh, and since the contributing guide requires sign off and allows to put it to the pull request comment I'll drop a line here:
`Signed-off-by: Sergey Shpikin <rkfg@rkfg.me>`

rkfg:
Removed obsolete settings and regrouped the rest. Seems that they need to be ported to PreferencesManager.

bmarty:
Hi @rkfg ,

Seem that your PR has been forgotten, sorry about that. Riot does contains a font size settings now, and the code has changed a lot since you've submitted the PR.

I'm afraid it will be hard to rebase it.

What do you want to do with this PR?

Benoît

rkfg:
Thanks for your attention. This patch is mostly obsolete, however I'd still like to have some bits of it implemented: the default media source and default compression settings. It would make posting photos so much more convenient. For now I always have to select the source (photo), take a picture, then select the same compression (medium). Having a preference for those would be nice and it's really easy to implement like I already did once. Other than that this PR can be closed safely.

Issue #1353, opened 2017-06-19
submit two riot apps to playstore as a poor man's multiaccount support


ara4n:
one for personal account; one for professional account, or similar, until we have proper multiaccount

jcgruenhage:
Another way to tackle this would be to give different builds different packages. 
Example:  

| Build | package |
| ---- | ---- |
| F-Droid Release | im.vector.riot.fdroid.release |
| F-Droid Debug | im.vector.riot.fdroid.debug |
| Play Store Release | im.vector.riot.googleplay.release |
| Play Store Debug | im.vector.riot.googleplay.debug |

Or something like that.

pganssle:
This is awful. If people want to go this route, I would think that they should use [one of the suggestions here](https://www.quora.com/How-do-I-install-the-same-app-multiple-times-on-my-Android-device) to clone the app on their own phone.

Adding a second version to the app store will just confuse people, and probably won't make for a particularly pleasant user experience anyway.

maxidor:
agreed with the horrible user experience. when telling people to just find and install Riot, they won't know which one to install...

Valodim:
I think this is actually a quite beautiful solution! (that's why I suggested it)

Having two apps in the play store covers the use case of connecting to two independent homeservers perfectly, and with minimal development effort. It's a straightforward user experience, the apps are separated in a way that is easy to understand and trust. Multiple accounts in the same app always feel iffy, e.g. if you want to connect to an internal business homeserver.

Patching apps to install them multiple times simply isn't a solution that works for most people.

Valodim:
I can see the argument about confusing users, so here's a fun suggestion: Call the second version "Riot Pro", give it a blue R icon rather than a green one, sell it for 1.99€, and make it very clear in the description that it's the same as the free one :)

Think about it: The target audience for this use case is people who would like to connect to two homeservers, one of them likely a company-hosted one. 1.99€ is pocket change so it won't bother the people who actually have this use case, it brings in a small amount of money, but at the same time it makes it perfectly clear which app to install for typical users. It's even a nice way to support the project with a very low barrier (everyone buys apps, not everyone has paypal), heck I'd go for that blue R.

maxidor:
@Valodim That's actually a very good idea

jcgruenhage:
People don't read descriptions and will assume that the free version is limited. Just an assumption, but still.

Mikaela:
[...]

PR #1552, opened 2017-08-31
Different application IDs for different configs

jcgruenhage:
A unique application ID based on buildType and productFlavor.
This has the advantage, that people can install dev builds alongside stable builds, either for trying out the dev build, or for having a stable backup, should the dev build break. 

This is also somewhat related to #1353, because this at least allows people to install the app 4 times, which should be enough for most people, and does so without having multiple apps in one store (which can be confusing).

### DIFF
--- a/vector/build.gradle
+++ b/vector/build.gradle
@@ -32,6 +32,7 @@ android {
 
     buildTypes {
         debug {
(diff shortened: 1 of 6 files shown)


jcgruenhage:
oh lol wrong branch

PR #1553, opened 2017-08-31
WIP: Different application IDs for different configs

jcgruenhage:
A unique application ID based on buildType and productFlavor.
This has the advantage, that people can install dev builds alongside stable builds, either for trying out the dev build, or for having a stable backup, should the dev build break. 

This is also somewhat related to #1353, because this at least allows people to install the app 4 times, which should be enough for most people, and does so without having multiple apps in one store (which can be confusing).

### DIFF
--- a/vector/build.gradle
+++ b/vector/build.gradle
@@ -10,7 +10,7 @@ android {
     }
 
     defaultConfig {
-        applicationId "im.vector"
+        applicationId "im.vector.alpha"
         minSdkVersion 16
         targetSdkVersion 25
         // use the version code
@@ -32,6 +32,7 @@ android {
 
     buildTypes {
         debug {
+            applicationIdSuffix ".dev"
             resValue "string", "git_revision", "\"${gitRevision()}\""
             resValue "string", "git_revision_date", "\"${gitRevisionDate()}\""
             resValue "string", "git_branch_name", "\"${gitBranchName()}\""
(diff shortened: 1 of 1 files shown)


jcgruenhage:
(sorry for the first pr to the wrong branch)

jcgruenhage:
This app has a provider, which makes this harder, as the package of that one would need to be changed as well.

jcgruenhage:
```
Execution failed for task ':vector:installAppfdroidDebug'.
> com.android.builder.testing.api.DeviceException: com.android.ddmlib.InstallException: Failed to finalize session : INSTALL_FAILED_CONFLICTING_PROVIDER: Package couldn't be installed in /data/app/im.vector.alpha.fdroid.dev-1: Can't install because provider name im.vector.VectorApp.provider (in package im.vector.alpha.fdroid.dev) is already used by im.vector.alpha
```

jcgruenhage:
After a little research, I found this: https://blog.grandcentrix.net/how-to-install-different-app-variants-on-one-android-device/

Got no time to implement this though, gotta study for exams.

ylecollen:
applicationIdSuffix ".dev" won't work because we will have some conflicts with the installed application.

we never implemented two versions (one dev and one release) to avoid having conflicts between pushes (and so on).

The froid version must keep the same value else i'm afraid the store update won't work.

jcgruenhage:
All those conflicts are resolvable though. And for the advantage it provides, it should be worth trying IMO.

The part with the F-Droid change, yes, that will break the update and force everybody to install that, and uninstall the other one, to continue to receive updates. That could be circumvented by discarding the .fdroid suffix, but at least the .dev suffix would be nice to have.

jcgruenhage:
This PR is stale so I'm going to close this.

PR #1647, opened 2017-10-06
Make use of jenkins BUILD_NUMBER in versionCode

krombel:
The fdroid client uses the versionCode to determine if updates are available.
If the jenkins-build-number (passed as variable BUILD_NUMBER in the env) would have an effect on the versionCode it would be possible that the fdoid-client notifies the users of the dev-build-repo of a new build.

In the current situation the new versionCodeProp would be set to 741660

Signed-Off-by: Matthias Kesler <krombel@krombel.de>


ylecollen:
https://github.com/vector-im/riot-android/commit/64ef115f62339ba1efba70e400451fdd370724eb
https://github.com/vector-im/riot-android/commit/0164dd3aeeb69f2d86beb4c1b67ce71da6aad314

Should help to track the build number

PR #1653, opened 2017-10-10
Compile against API 26 (Android Oreo).

### LINKED ISSUE #1633: Update targetSdkVersion to android-26
Hi!

 Android Oreo is out there, and I was wondering if there is any problem with updating the targetSdkVersion to android-26.

 Thank you very much

rafagalvez:
This closes #1633. Rather than targetSdkVersion, we change compileSdkVersion (see [difference](https://stackoverflow.com/questions/26694108/what-is-the-difference-between-compilesdkversion-and-targetsdkversion)

Changing targetSdkVersion also makes sense, but since there are no automated test suite, I cannot test that everything works fine with that. The main benefit would apparently be that the visuals may be prettier.


rafagalvez:
Deleted the commit in develop, branched out.

rafagalvez:
Reopened in #1654

PR #1654, opened 2017-10-10
Compile against API 26 (Android Oreo).

### LINKED ISSUE #1633: Update targetSdkVersion to android-26

rafagalvez:
This closes #1633. Rather than targetSdkVersion, we change compileSdkVersion (see difference

Changing targetSdkVersion also makes sense, but since there are no automated test suite, I cannot test that everything works fine with that. The main benefit would apparently be that the visuals may be prettier.

### DIFF
--- a/vector/build.gradle
+++ b/vector/build.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 25
+    compileSdkVersion 26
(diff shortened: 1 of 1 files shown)


ylecollen:
I update also the build tools and the compat libs

PR #1655, opened 2017-10-10
Update Android plugin for Gradle.

rafagalvez:
This enables "Instant Run". Citing https://developer.android.com/studio/run/index.html#instant-run:

In Android Studio 2.3 and higher, Instant Run significantly reduces the time it takes to update your app with code and resource changes. After deploying your app to a target device running Android 5.0 (API level 21) or higher, you can click Apply Changes to push certain code and resource changes to your running app without building a new APK—and, in some cases, without even restarting the current activity.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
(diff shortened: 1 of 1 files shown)


ylecollen:
Was done with android o update

Issue #1663, opened 2017-10-12
Quoting uses more newlines than needed


nvbln:
As an example, when I type:

```
> test

Test
```

The following will happen:

![file1507809859542](https://user-images.githubusercontent.com/20875906/31495142-aea0eafe-af56-11e7-948e-81ecf409d2ec.png)

As you can see in the image, instead of one newline, it will create three newlines (one extra inside the quote and two outside the quote).

bmarty:
this issue is fixed now

Issue #166, opened 2016-05-31
We choke comically on markdown tables


matrixbot:
Created by @ matthew:matrix.org.

See https://vector.im/develop/#/room/!cURbafjkfsMDVwdRDQ:matrix.org/$1464713154349iypGR:jki.re

ylecollen:
Patched in https://github.com/matrix-org/matrix-android-sdk/commit/2e0fd32dd82e611a38b8da787fab565526ff4891

PR #1735, opened 2017-11-07
Remove html text replacer

nvbln:
The replace function that replaces newlines with break tags does not seem to have a good use and breaks other things (see issue #1663 for example). I think the replace function was meant to improve the table layout, but as markdown tables are not supported by commonmark, it only causes formatting issues right now.


Matrixcoffee:
I couldn't find a reason for the <br/> replacement to be there. Not to say there isn't one, but it seems to have gone undocumented and is now lost to history.

The lines (of which there are now two) were originally added in SDK commit dc8498b57eec762f48f0b8826914141151960959, which message merely says "Fix the \n issue in markdown".

I looked for contemporary issues in github, but found none that applied.

The tables comment seems unrelated and was added later along with some code. See #166.

Let's remove this line and find out what happens on various phones and versions of Android. Worst case, we can simply put it back.

PR #1762, opened 2017-11-18
Feature: send voice message as attachment

harish2704:
We know that Riot-Android do not have "push to talk" kind feature,
This PR provides a alternative way or work-around for this issue using a simple solution.
Ie, send a voice message attachment using system-sound-recorder application.

The work flow is as follows
* User clicks on 'attachment button"
* in addition to "send-fiels" & "take photo or video", one more additional choice called "send voice" is added
* when user clicks on "send voice", system voice recorder application will get invoked and user can record voice using it.
* Finally, resultant file is sent as file attachment.

When we implement built-in Push-to-talk button and proper voice encoding , we can replace this feature with new PTT button. Till that time I hope this will be helpful to lot of normal users

Signed-off-by: Harish.K <harish2704@gmail.com>


harish2704:
screenshot_20171118-195541

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -721,6 +725,8 @@ public void onItemClick(IconAndTextDialogFragment dialogFragment, int position)
 
                             if (selectedVal == R.string.option_send_files) {
                                 VectorRoomActivity.this.launchFileSelectionIntent();
+                            } else if (selectedVal == R.string.option_send_voice ) {
+                                VectorRoomActivity.this.launchAudioRecorderIntent();
```
ylecollen: audio permission request must be check too

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -2163,6 +2169,16 @@ private void launchInvitePeople() {
         }
     }
 
+    /**
+     * Launch audio recorder intent
+     */
+    @SuppressLint("NewApi")
```
ylecollen: Why did you add this "SuppressLint" ?

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -2163,6 +2169,16 @@ private void launchInvitePeople() {
...
+    private void launchAudioRecorderIntent() {
+        enableActionBarHeader(HIDE_ACTION_BAR_HEADER);
+        Intent fileIntent = new Intent( MediaStore.Audio.Media.RECORD_SOUND_ACTION );
+        startActivityForResult(fileIntent, REQUEST_FILES_REQUEST_CODE);
```
ylecollen: RECORD_SOUND_ACTION seems optional (https://stackoverflow.com/questions/15546051/android-mediastore-audio-media-record-sound-action-no-activity-found-to-handle).
Notice, we might have the same issue for the other entry (camera...)

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
harish2704: Sorry it happened during copy-paste. Will remove

bmarty:
Hi @harish2704,

Thanks for this PR. It will be merged soon :)
This feature will be added to the Lab feature for the moment.
Could you please sign-off it as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off

Benoît

harish2704:
Hi @bmarty 
I just pushed a new commit which removes some unwanted comments ( as specified by @ylecollen )
I have added a Sign-off message in that commit. Hope it will resolve the problem

BloodyIron:
I'm not seeing how to do this intuitively from the UX. And I also cannot figure out how to do this at all. How exactly was this implemented from a UX perspective? D:

bmarty:
Whatsapp has a good UX for this. But for now this is just a quick implementation to send audio message. In the future, voice message qhould be handled by the matrix protocol and implemented with a good UX on the Riot clients

BloodyIron:
Okay, but I don't see how (as in, I literally don't see this) I send a voice recording to a channel. I believe the feature was added, but I have no idea how to use it... How does one use it? This is half asking how, half pointing out a UX problem.

BloodyIron:
Hmmm only seem to be able to download with the desktop version, not just play in-line. Is that functionality coming? (play the recording I posted in a channel)

irregularisnotrandom:
I am.gettting certificate error on android when i enable this feature in the labs. After i enable and reopen the app it tells.me there is an.issue with my LE cert. I wipe the data and readd my account no cert issue. I tried this 2 times already and both times i get this cert issue. Since last fresh start without this experimental.issue i.never had and cert issue

PR #1780, opened 2017-11-26
Start refactoring the room layout

kaiyou:
First disclaimer: this is probably not ready to merge yet, I am merely creating the pull request to keep track on the work and benefit from Jenkins builds.

This includes some changes to the general room UI, in order to get a simpler look&feel for unexperimented users and a generally prettier output:
- the message placeholder is generic, to avoid creating a false impression of non-e2ee rooms being completely insecure,
- the room encryption state was moved to the header toolbar, next to the room title
- the search button was removed from the header toolbar to save some space
- the room topic was removed in the header toolbar, and replaced with a dynamic notification toolbar, that displays active users, typing notification and server connection issues
- unsent message errors are now displayed as a standard view with buttons, instead of a spantext with clickable links.


kaiyou:
Here is current output:

ara4n:
hm, this is interesting work - thanks for the PR.  It's certainly addressing some valid concerns, although there's a bit of a collision between the proposed UI and the route that the app's designers (yes, Riot's UX was originally designed by professionals) had in mind.  For instance, the idea for getting rid of the annoying status bar is to show it on demand (a bit like Riot/Web does these days) - and for typing notifications to be shown as placeholder messages - like iMessage does.

However, the screenshots do look nice here.  I'm a bit worried that putting typing up in the title might be hiding it too much (whereas the typing notifs conceptually might be better at the bottom as they're showing the most recent thing happening in the room).

I think the best is for you to flesh it out to match your mental image, and then we and others can play with it and see how it feels.  But I should warn you that we can't guarantee merging major cosmetic changes to the app, given there is (in theory at least) a bigger story on progress with the design.

kaiyou:
Thank you for the feedback. is there anywhere you documented the overall roadmap in terms of UI/UX except for the various issues?

Generally speaking, even though a pr is not the proper place to discuss this, I think on mobile, most people expect: 1. the layout to be as static as possible (i.e. do not change the structure while using the app) 2. contextual info to be displayed in a single place, usually at the top.

I just pushed a couple more commits, including the scrolldown arrow which used to be part of the notification bar. It is far from perfect but should be more or less functionally equivalent now.

kaiyou:
I guess this can be closed given the current efforts on the ui/ux

bmarty:
Hi @kaiyou ,
What is the status of your PR? You're suggesting to close it?
Benoit

kaiyou:
Hi @bmarty, as I commented already, I guess we can close this. The effort would need to be spent again given the changes in the UI, also it is not currently clear where Riot Andoid is going in terms of UI/UX, so I guess I should wait until a proper roadmap is paved until I contribute again, so that suggestions are compatible with the overall strategy :)

PR #1796, opened 2017-12-07
Update Ukrainian translation

elsieholmes:
This PR brings a translation of a brand new bug_report_activity


elsieholmes:
Really? This PR is half of a month old...

PR #1811, opened 2017-12-16
Fixes #145: improve the code display (double and triple backticks)

### LINKED ISSUE #145: improve the code display (double and triple backticks)
from android chat

it would be nice if single and tripple backtick blocks (ie: code blocks) had a different background color,
and tripple backtick blocks didn't get wrapped, instead had a scrollbar as they do in vector web

Morever, the webclient formats the code blocks with another way than backticks.
I assume it is the third case ie the formatted code text should also be displayed with another background.

pacamara:
Fixes #145: "improve the code display (double and triple backticks)"

Introduces a new ROW_TYPE_CODE for messages that contain a mixture of fenced (i.e. 3 backtick) code blocks and text sections. The text sections can contain in-line (unfenced) code blocks. Each of the fenced code blocks is in its own HorizontalScrollView and independently scrollable.

This is done via a new layout adapter_item_vector_message_code.xml, which can contain a mixture of  adapter_item_vector_message_code_block.xml and adapter_item_vector_message_code_text.xml layouts. 

A different background colour for fenced code blocks can be specified using the new attribute fenced_code_block_background_color. The current colour is by way of example only, and is intended to be changed by designers.

issue145_fixed_screenie_postreview

Screenshot showing a single message containing:
* two fenced (3 backtick) code blocks, one of which has been scrolled
* a plain text section containing inline code
* plain text sections with no code

### DIFF
--- a/vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
+++ b/vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
@@ -75,6 +75,9 @@ import im.vector.widgets.WidgetsManager;
 class VectorMessagesAdapterHelper {
     private static final String LOG_TAG = VectorMessagesAdapterHelper.class.getSimpleName();
 
+    /** Enable multiline mode, split on ``` at start of line and capture delimiters */
+    private static final Pattern FENCED_CODE_BLOCK_PATTERN = Pattern.compile("(?m)(?<=^```)|(?=^```)");
+
     private IMessagesAdapterActionsListener mEventsListener;
     private final MXSession mSession;
     private final Context mContext;
@@ -552,38 +555,26 @@ class VectorMessagesAdapterHelper {
         }
     }
 
-    /** Determine how many backticks bracket a message, for code block
-     * handling (issue 145) */
-    public int getTickCount(final Message message) {
-        final String mb = message.body;
-        if (mb.startsWith("```") && mb.endsWith("```")) {
-            return 3;
-        }
-        else if (mb.startsWith("``") && mb.endsWith("``")) {
-            return 2;
-        }
-        else if (mb.startsWith("`") && mb.endsWith("`")) {
-            return 1;
-        }
-        else {
-            return 0;
-        }
+    /** Determine if the message contains any code blocks (issue 145). */
+    public boolean containsFencedCodeBlocks(final Message message) {
+        final String[] blocks = getFencedCodeBlocks(message);
+        return (blocks.length > 1);
     }
 
-    /** Handling for ROW_TYPE_CODE (issue 145) */
-    void highlightCode(final TextView textView, final Spannable text, final int tickCount) {
+    /** Result includes the ``` delimiters as separate entries (issue 145). */
+    public String[] getFencedCodeBlocks(final Message message) {
+        return FENCED_CODE_BLOCK_PATTERN.split(message.body);
+    }
+
+    /** Issue 145 */
+    void highlightFencedCode(final TextView textView, final Spannable text) {
         // sanity check
         if (null == textView) {
             return;
         }
 
-        final int background = (tickCount==1) ?
-                ThemeUtils.getColor(mContext, R.attr.code_block_1_background_color) :
-                (tickCount==2) ?
-                ThemeUtils.getColor(mContext, R.attr.code_block_2_background_color) :
-                ThemeUtils.getColor(mContext, R.attr.code_block_3_background_color);
+        final int background = ThemeUtils.getColor(mContext, R.attr.fenced_code_block_background_color);
         textView.setBackgroundColor(background);
-
         textView.setText(text);
 
         if (null != mLinkMovementMethod) {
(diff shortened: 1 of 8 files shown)


vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
```diff
@@ -552,6 +552,45 @@ public void onClick(View view) {
         }
     }
 
+    /** Determine how many backticks bracket a message, for code block
+     * handling (issue 145) */
+    public int getTickCount(final Message message) {
+        final String mb = message.body;
+        if (mb.startsWith("```") && mb.endsWith("```")) {
```
ylecollen: 1 - what is the plan is the message contains several blocks of code ?
you only manage 1 block code by message.

2 - the markdown might be disabled.

vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
```diff
@@ -552,6 +552,45 @@ public void onClick(View view) {
...
+            return 3;
+        }
+        else if (mb.startsWith("``") && mb.endsWith("``")) {
+            return 2;
+        }
+        else if (mb.startsWith("`") && mb.endsWith("`")) {
+            return 1;
+        }
+        else {
+            return 0;
```
ylecollen: give a name for ``, ``, ` block

vector/src/main/res/layout/adapter_item_vector_message_code.xml
```diff
@@ -0,0 +1,121 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ylecollen: i suspect it is the same layout than the text message

JonasDralle:
This fixes the problem of newlines being ignored, right?

pacamara:
@JonasDralle Yes, eg in the HelloWorld code example in the screenshot, code is not wrapped and newlines are not ignored.  
@ylecollen Thanks for the review!

vector/src/main/res/layout/adapter_item_vector_message_code.xml
pacamara: It's different from adapter_item_vector_search_message_text_emote_notice.xml in three ways:  
i) wraps the TextView in a HorizontalScrollView
ii) monospace font
iii) no auto-linking of URLs

vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
pacamara: 1. Ah, I didn't know this was a requirement. In that case, if the message contains one or more code blocks, use the adapter_item_vector_message_code.xml  layout, and use spans to set the colour of each block.
2. Will fix this.

pacamara: Issues 1 and 2 both fixed now.

vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
pacamara: In my implementation of your review, this function no longer exists.

pacamara:
@ylecollen @JonasDralle  I've implemented @ylecollen 's code review and pushed it as a separate commit. Will squash to a single commit when it's time to merge. A single message can now contain multiple fenced (3 backtick) code blocks each of which lives in its own independently scrollable view, as well as text sections which can contain inline code blocks. It also respects the markdown-enabled config setting now.

(I've removed support for a separate colour for inline code blocks since this is already controlled by an existing Riot attribute. Please let me know if you sitll want support for a separate colour).

ylecollen:
LGTM

ylecollen:
@pacamara 
could you provide us more info we add you in the contributors list ?

https://github.com/vector-im/riot-android/blob/master/AUTHORS.rst

pacamara:
@ylecollen Thanks! :)  Contributor info is:
Joseph Gittings <joe.spbn at gmail dot com>

It says in Status Open Bounty, "payout pending maintainer confirmation", so my payout is still pending, right? I have updated my SNT receiving address in SOB to the correct value.

ylecollen:
@pacamara 
it does not fill all the cases : for example, the https://imgur.com/3Hb4hgh is not fixed.
ie if you send a block of code from the webclient, the formatted body could contain the <pre><code>.. tags.

ylecollen:
improve with
https://github.com/vector-im/riot-android/commits/develop

but it does not fix all the cases.

pacamara:
@ylecollen Hi! Was chatting about this with @lampholder yesterday and we came up with these two to-do items to address the above concerns:

a) apply markdown-enabled setting at message send time, not display time.
b)  look for <code> and <pre><code> blocks in message. formatted_body rather than backticks in message. body, and split the blocks[] array on those.

Point (b) seems a simple enough change and would address compatibility between the web and android clients. I'm happy to do both changes or would you rather?

ylecollen:
@pacamara 
i only patched your code until you have time to fix everything properly.
i've got some comments that the code block height and its background colour.

pacamara:
@ylecollen Cool, I'll get on this and build on your changes then.  :+1:

pacamara:
@ylecollen  OK done! See https://github.com/vector-im/riot-android/pull/1878  :beers:

PR #1878, opened 2018-01-10
issue 145: enhancements to previous fixes for code block display

pacamara:
* Builds on #1811 and https://github.com/vector-im/riot-android/commit/545d46fc67a76e4b334be356f5b9cc9130cbdaa4
* For fenced code blocks, split on `<pre><code>...</code></pre>` in message.formatted_body, not triple backticks in message,body
* Tweak wording of "Disable markdown formatting" menu item to clarify it refers to sending messages, not displaying existing messages.

Note: ``` for fenced code blocks are still required to be on a separate line: this is due to the parsing code in the Matrix SDK. Also, this is the behaviour of the Riot web client. So, haven't attempted to change this.

### DIFF
--- a/vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
+++ b/vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
@@ -1209,12 +1209,15 @@ public class VectorMessagesAdapter extends AbstractMessagesAdapter {
         for (final String block:blocks) {
             if (block.startsWith(START_FB) && block.endsWith(END_FB)) {
                 // Fenced block
-                final String stripped = block.substring(START_FB.length(), block.length()-END_FB.length());
-                final SpannableString threeTickBlock = new SpannableString(stripped);
+                final String minusTags = block.substring(START_FB.length(), block.length()-END_FB.length());
                 final View blockView = mLayoutInflater.inflate(R.layout.adapter_item_vector_message_code_block, null);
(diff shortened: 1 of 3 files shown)


ylecollen:
The main work should be done in org.matrix.androidsdk.view.HtmlTagHandler.
This class is the custom text renderer when Html.fromHtml is used.
The HTML tags are parsed and managed to build a formatted text.

pacamara:
@ylecollen What other work needs doing? The requirement for the triple backticks to have their own line is the same as in the web client, so that presumably does not need changing. The formatted_text built by HtmlTagHandler is already correctly handling fenced code blocks. :smile:

ylecollen:
@pacamara 
your code does not manage \<code\> and \<pre\>\<code\> tags properly.

-> If the android sends an unformatted triple backticks code, the HTML tags are \<code\> : your code won't work.
-> the webclient sends \<pre\>\<code\> blocks

pacamara:
I don't think so! The android client sends `<pre><code>` for triple backticks, as long as the triple backticks are on a separate line. [Screenshot from the android debugger here.](https://ibb.co/gjr4rm)

ylecollen:
LGTM

pacamara:
Cheers! Just noticed one small further change needed, now we're calling highlightPattern for the fenced blocks we need to protect the line breaks and spaces. See #1882

PR #1880, opened 2018-01-11
fix #1460

### LINKED ISSUE #1460: Broken regular GSM-call when Riot calling
I'm talking on regular GSM on a smartphone. When a call arrives in Riot, a notification of the call appears, and at the other end it stops hearing me. Someone should be the main one when working with a microphone. I think that the GSM call should be the main one. Is it possible to do something about it? Maybe if I already talk on the phone, answer that I'm busy automatically.

dva-re:
fix issue #1460 
Hangup incoming call when GSM call in progress


ylecollen:
LGTM

PR #1955, opened 2018-01-29
add_exit_button

ylecollen:
Add an exit button in the home sliding menu


bmarty:
I do not understand why people seems to be happy with this feature. Calling System.exit(0) in an Android application is not a recommended practice.
I may missed something...

PR #2003, opened 2018-02-22
Fix notifications being unswipable on fdroid builds

### LINKED ISSUE #2001: Can't swipe away notifications since v0.8.1
After updating to the F-Droid 0.8.1 version I can't dismiss/ignore notifications by swiping them away any more. That means if I don't want to read something now the notification just has to sit there, occupying notification space until I have time to deal with it.

Related to #1671 but this is about the action of ignoring a notified message for the moment instead of marking it as read.

jgkamat:
The notification logic was setting normal incoming notifications as 'foreground notifications' which are un-swipable by design (as they are the ones that keep the app open). This PR tries to re-factor this, so notifications that aren't part of the service (message based ones) aren't put in the foreground, and don't interfere with the actual foreground notification.

This also switches the 'caller notifications' to be set as un-swipable using flags, rather than using them as the foreground service (which is hacky, in my opinion). As a side effect of this, the forgeround service notification will still be visible when notifications come in (which is in line with other apps that use this behavior, I think).

I haven't done anything with android notifications before, so let me know if anything seems wrong.

This closes #2001.

```
Signed-off-by: Jay Kamat <jaygkamat@gmail.com>
```


jgkamat:
I also branched off a little earlier than develop tip because #1999 was causing some compiler errors and I didn't want to look into it too deeply.

vector/src/main/java/im/vector/services/EventStreamService.java
```diff
@@ -1640,9 +1633,11 @@ else if (null == CallsManager.getSharedInstance().getActiveCall()) {
      * @param callId  the callId
      */
     public void displayCallInProgressNotification(MXSession session, Room room, String callId) {
+        final NotificationManagerCompat nm = NotificationManagerCompat.from(EventStreamService.this);
```
ylecollen: the call notification must be in foreground to keep the call alive.
FLAG_NO_CLEAR is not enough.

vector/src/main/java/im/vector/services/EventStreamService.java
```diff
@@ -1598,6 +1590,7 @@ private boolean refreshNotifiedMessagesList() {
      */
     public void displayIncomingCallNotification(MXSession session, Room room, Event event, String callId, BingRule bingRule) {
         Log.d(LOG_TAG, "displayIncomingCallNotification : " + callId + " in " + room.getRoomId());
+        final NotificationManagerCompat nm = NotificationManagerCompat.from(EventStreamService.this);
```
ylecollen: the call notification must be in foreground to keep the call alive.
FLAG_NO_CLEAR is not enough.

vector/src/main/java/im/vector/services/EventStreamService.java
```diff
@@ -853,10 +854,10 @@ public void refreshStatusNotification() {
         if (mNotificationState == NotificationState.NONE) {
             Notification notification = NotificationUtils.buildMessageNotification(getApplicationContext(), true);
 
+            final NotificationManagerCompat nm = NotificationManagerCompat.from(EventStreamService.this);
             if (null != notification) {
                 mNotificationState = NotificationState.DISPLAYING_EVENTS_NOTIFICATIONS;
-                startForeground(NOTIFICATION_ID, notification);
-                mIsForeground = true;
+                nm.notify(NOTIFICATION_ID, notification);
```
ylecollen: This notification is used to keep alive the application alive during the application launch. (it should be updated after the first sync request).

It is used by the Google play AND the  Fdroid versions.

Without this foreground notifications, the application might be killed during its launch.
for the google play version, some notifications might not be triggered.
for the fdroid version, the application might be killed whereas it was starting on boot.

vector/src/main/java/im/vector/services/EventStreamService.java
```diff
@@ -873,15 +874,15 @@ public void refreshStatusNotification() {
             Log.d(LOG_TAG, "## refreshStatusNotification : put the service in foreground because of an initial sync " + mNotificationState);
 
             if (mNotificationState != NotificationState.INITIAL_SYNCING) {
-                startForeground(NOTIFICATION_ID, buildForegroundServiceNotification(getString(R.string.notification_sync_in_progress)));
+                startForeground(NOTIFICATION_ID_PERSISTENT, buildForegroundServiceNotification(getString(R.string.notification_sync_in_progress)));
```
ylecollen: This notification should use the event notification id.
It replaces it during the application launch (see my previous comment).

vector/src/main/java/im/vector/services/EventStreamService.java
```diff
@@ -1352,16 +1353,7 @@ public void run() {
...
-                                    mIsForeground = true;
-                                    startForeground(NOTIFICATION_ID, notif);
-                                } else {
-                                    if (mIsForeground) {
-                                        stopForeground(true);
-                                        mIsForeground = false;
-                                    }
-                                    nm.notify(NOTIFICATION_ID, notif);
-                                }
+                                nm.notify(NOTIFICATION_ID, notif);
```
ylecollen: the plan was to merge the both notifications in a single one
1- many fdroid users complained about it
2- on google play version, during the GCM registration, the service must be put in foreground to trigger notifications.

As yous said in the android room, the froid users are used to have both notifications (at least some of them).
But, we have to have a "simpler" behaviour for the google play users. Many of them already complained about this "listen for events" notification.

ylecollen:
@jgkamat 
See my comments on your PR,
Your fix only improves/fixes the fdroid notifications management but it breaks the google play one.

You should use mGcmRegistrationManager.useGCM() to detect the fdroid version.

if (fdroid) -> have a dedicated "listen for events" foreground notification id (the event id one might also be in foreground as explained in the code review)

if (fdroid) -> do not notify on foreground.

I let you try to implement the fdroid behaviour without breaking the google play one.

As you understand by now, it is not as simple as you thought.

vector/src/main/java/im/vector/services/EventStreamService.java
jgkamat: Setting this to a foreground notification causes the first notification when starting the app to become unswipable on both google play and fdroid. However, I don't want to do too big of a restructure, so I'll just restore the old behavior.

vector/src/main/java/im/vector/services/EventStreamService.java
jgkamat: To fix this, I made the events notification come in over the PERSISTENT channel instead. Let me know if it still seems wrong.

vector/src/main/java/im/vector/services/EventStreamService.java
jgkamat: FWIW, I think the previous behavior is less 'simple' as it leads to people thinking the app is broken (since it produces un-swipable notifications). 

As for point 2, shouldn't the 'Listen for events' notification come up if GCM registration fails regardless of this section? Ideally, people on google play should never see the 'listening for events' notification, so I don't think it's worth it to make a special case for if a user is on google play but dosen't have GCM. 

If you think otherwise though, I can restore the previous behavior on google play when GCM fails, what do you think?

jgkamat:
This is slightly a repeat from the line comments but:

> if (fdroid) -> do not notify on foreground.

This section seems a bit wrong to me, as it would cause notifications to become unswipable on google play if GCM fails. I think the correct behavior is to have the same behavior as fdroid if GCM registration fails (have a persistent forground notification). I think doing it in another way leads to confusion from people unable to swipe away notifications.

If you really think I should make a special case for google play when GCM fails, let me know, and I'll go ahead and implement it.

ylecollen:
@jgkamat 

 If the GCM registration fails, you have to choose between having two notifications (what the google play users won't understand) and having an unswipable notification.
 There is not good choice.

 Moreover, since android 8,
->  it is not anymore possible to hide a foreground notification
-> the service must be put in foreground to perform an unique sync when a push is received (before being put in normal state)
   Background services are killed < 15 seconds when the application is put in background.
-> on android 8 devices; it might trigger a "synchronizing" notification if there is no notifications.

  
  To avoid this behaviour, you can grant the "run in background" permission : the application behaviour will be the one you expect.
  i already added a dialog to request it : **ALL** users rejected it even after explaining the reason.

Issue #1971, opened 2018-02-04
Re-enable rageshake?


rkfg:
So I disabled the rageshake reporter a while ago and now want to enable it back to report the bug with a screenshot. I suppose if I use the "Bug Report" from the side menu it doesn't take a screenshot of the room I'm currently in because I have to move out of it to reach the menu. How do I re-enable shake detection? Couldn't find it in the options and don't want to reinstall the app just because of that.

manuroe:
FTR, there is a "Rage shake to report bug" button on riot-ios. We could have it on riot-android too.

PR #2008, opened 2018-02-22
SETTINGS : Re-enable rageshake

Tyuoli:
Use PreferenceManager to save rageshake mode instead of using the sdk
(https://github.com/vector-im/riot-android/issues/1971)
(https://github.com/matrix-org/matrix-android-sdk/pull/236)


Tyuoli:
I took your remarks into account so that the value of rageshake is well recorded in the SharedPreferences of the application

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.java
```diff
@@ -207,8 +208,9 @@ public void onAccountInfoUpdate(MyUser myUser) {
     private EditTextPreference mSyncRequestTimeoutPreference;
     private EditTextPreference mSyncRequestDelayPreference;
     private PreferenceCategory mLabsCategory;
-
     private PreferenceCategory mGroupsFlairCategory;
+    // tell if the rageshake mode is enabled
+    private boolean mIsUsedRageShake;
```
manuroe: This ivar seems useless to me.

Issue #1985, opened 2018-02-09
room aliases in topic not click able


t3chguy:
only the server_name is linkified

PR #2013, opened 2018-02-23
VectorRoomHeader : clickable

Tyuoli:
(https://github.com/vector-im/riot-android/issues/1985)

### DIFF
--- a/vector/src/main/java/im/vector/activity/VectorRoomActivity.java
+++ b/vector/src/main/java/im/vector/activity/VectorRoomActivity.java
@@ -2285,7 +2285,7 @@ public class VectorRoomActivity extends MXCActionBarActivity implements MatrixMe
         values.put(MediaStore.Images.Media.TITLE, CAMERA_VALUE_TITLE + formatter.format(date));
         // The Galaxy S not only requires the name of the file to output the image to, but will also not
         // set the mime type of the picture it just took (!!!). We assume that the Galaxy S takes image/jpegs
-        // so the attachment uploader doesn't fremakeLinkClickableak out about there being no mimetype in the content database.
+        // so the attachment uploader doesn't freak out about there being no mimetype in the content database.
         values.put(MediaStore.Images.Media.MIME_TYPE, "image/jpeg");
         Uri dummyUri = null;
         try {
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -2282,7 +2285,7 @@ private void launchNativeCamera() {
         values.put(MediaStore.Images.Media.TITLE, CAMERA_VALUE_TITLE + formatter.format(date));
         // The Galaxy S not only requires the name of the file to output the image to, but will also not
         // set the mime type of the picture it just took (!!!). We assume that the Galaxy S takes image/jpegs
-        // so the attachment uploader doesn't freak out about there being no mimetype in the content database.
+        // so the attachment uploader doesn't fremakeLinkClickableak out about there being no mimetype in the content database.
```
manuroe: ?

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -3006,11 +3009,50 @@ private void updateRoomHeaderTopic() {
...
+                }
             }
         }
     }
 
+    /**
+     * Trap the clicked URL.
+     *
+     * @param strBuilder    the input string
+     * @param span          the URL
```
manuroe: Comment for `roomAlias` is missing.

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -3006,11 +3009,50 @@ private void updateRoomHeaderTopic() {
...
+     */
+    public void makeLinkClickable(SpannableStringBuilder strBuilder, final URLSpan span, final String roomAlias) {
```
manuroe: `roomAlias` can be replaced by a more generic term like ... `value` (better naming is welcome).
The idea is room alias is one possible value but it can be also a room id, a user id, a group id, etc...

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
Tyuoli: it's a simple issue of copy and paste.

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
Tyuoli: I agree for use generic term value instead of roomAlias because value can be a roomId or else.

PR #2036, opened 2018-03-04
Hide notification area.

ma1uta:
There is a setting to hide notification area because it's annoying me sometimes when waste space on the mobile screen.
There are 3 options: always, never, when empty.
Default value is "never".
Also added translation using Google translate.

Signed-off-by: Anatoly Sablin <sablintolya@gmail.com>


Matrixcoffee:
By the way, I think "always" is a bad idea, because it hides send errors, and prevents Riot from asking if you want to send it again. It would be nice if it could be set to not open for typing notifications though.

I can't think of any good names for those options though. Perhaps it makes more sense to invert the setting, renaming it to "Show notification area", with options "errors only", "messages", "always".

ma1uta:
I agree with you that "always" is bad idea. I will change it to the "errors only".

About an inverting the settings: I can do it, no problem.

ma1uta:
Done. I reverted the preference from "Hide notification area" to "Show notification area" with three options: "always", "only errors" and "messages and errors".

vector/src/main/res/values/donottranslate.xml
```diff
@@ -36,4 +36,16 @@
         <item>@string/dark_theme</item>
         <item>@string/black_them</item>
     </string-array>
+
+    <string-array name="notification_area_show_codes">
+        <item>only_errors</item>
```
bmarty: order should be:
1. always (the default value)
2. message and errors
3. only_errors

vector/src/main/res/values/donottranslate.xml
```diff
@@ -36,4 +36,16 @@
...
+        <item>messages_and_errors</item>
+        <item>always</item>
+    </string-array>
+
+    <string-array name="notification_area_show_description">
+        <item>@string/notification_area_show_only_errors</item>
```
bmarty: also change the order here

vector/src/main/java/im/vector/util/NotificationAreaUtils.java
```diff
@@ -0,0 +1,62 @@
...
+        SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(context);
+        return ALWAYS.equals(sp.getString(NOTIFICATION_AREA_SHOW_KEY, ALWAYS));
+    }
+
+    /**
+     * Sets default value if nothing specified.
+     *
+     * @param context the context
+     */
+    protected static void setDefault(Context context) {
```
bmarty: this method is not necessary, please remove it

vector/src/main/java/im/vector/util/NotificationAreaUtils.java
```diff
@@ -0,0 +1,62 @@
...
+        return ONLY_ERRORS.equals(sp.getString(NOTIFICATION_AREA_SHOW_KEY, ALWAYS));
+    }
+
+    /**
+     * Determine to show or hide the notification area.
+     *
+     * @param context the context
+     * @return {@code true} if invisible, else {@code false} for gone
+     */
+    public static boolean invisibleOrGone(Context context) {
```
bmarty: this method name is not clear. Should start with a verb. Comment is not clear as well

vector/src/main/java/im/vector/util/NotificationAreaUtils.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Check is notification area always hide.
+     * <p/>
+     * Used for optimization.
+     *
+     * @param context the context
+     * @return {@code true} if always hide, else {@code false}
+     */
+    public static boolean onlyErrors(Context context) {
```
bmarty: method name should start with a verb.

vector/src/main/java/im/vector/util/NotificationAreaUtils.java
```diff
@@ -0,0 +1,62 @@
...
+    public static final String NOTIFICATION_AREA_SHOW_KEY = "NOTIFICATION_AREA_SHOW_KEY";
+
+    /**
+     * Notification area hide preference values.
+     */
+    private static final String ONLY_ERRORS = "only_errors";
+    private static final String ALWAYS = "always";
+
+    /**
+     * Check is notification area always hide.
```
bmarty: This comment is not clear (update after only_error new value)

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -2686,6 +2694,17 @@ public void onClick(View v) {
         if (null != mUseMatrixAppsMenuItem) {
             mUseMatrixAppsMenuItem.setVisible(TextUtils.isEmpty(mEventId) && (null == sRoomPreviewData) && PreferencesManager.useMatrixApps(this));
         }
+
+        if (NotificationAreaUtils.onlyErrors(this)) {
+            int view = R.drawable.error == iconId ? View.VISIBLE : View.GONE;
+            mNotificationsArea.setVisibility(view);
+            findViewById(R.id.room_notification_separator).setVisibility(view);
```
bmarty: please bind this view with @BindView, to avoid to find it with findViewById each time you want to use it

PR #2050, opened 2018-03-07
Use the RoomUtils.isDirectChat method to test if a room is direct

Tyuoli:


vector/src/main/java/im/vector/adapters/VectorRoomSummaryAdapter.java
```diff
@@ -810,7 +806,7 @@ public View getChildView(int groupPosition, int childPosition, boolean isLastChi
         }
 
         if (null != childRoom) {
-            directChatIcon.setVisibility(mDirectChatRoomIdsList.indexOf(childRoom.getRoomId()) < 0 ? View.GONE : View.VISIBLE);
+            directChatIcon.setVisibility(!RoomUtils.isDirectChat(mMxSession, childRoom.getRoomId()) ? View.GONE : View.VISIBLE);
```
giomfo: Can you please remove the `!` in the condition and inverse the View.GONE, View.VISIBLE?

PR #2055, opened 2018-03-07
Fix `R.plurals.directory_search_rooms_for` use

giomfo:
+ code cleaning

### DIFF
--- a/vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
+++ b/vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
@@ -171,10 +171,12 @@ class VectorMessagesAdapterHelper {
         if (null != senderTextView) {
             Event event = row.getEvent();
 
+            // Hide the group flair by default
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
```diff
@@ -171,10 +171,12 @@ public void setSenderValue(View convertView, MessageRow row, boolean isMergedVie
         if (null != senderTextView) {
             Event event = row.getEvent();
 
+            // Hide the group flair by default
+            groupFlairView.setVisibility(View.GONE);
+            groupFlairView.setTag(null);
+
```
ylecollen: you might have flair icon flickerings on low end devices...

giomfo: @ylecollen I did some tests without observing flickering. I saw some delay in the flair display sometimes, but it was already the case before my changes. So I will keep my changes FTM, thank you to have warned me.

Issue #1921, opened 2018-01-20
Embedded images must be restricted to mxc: URIs


Matrixcoffee:
Right now img tags are allowed arbitrary linking, which is a security issue!

ylecollen:
https://github.com/vector-im/riot-meta/issues/133

ara4n:
yes, this is a security issue - and has been mistriaged. @giomfo presumably this is a trivial fix?

manuroe:
The security regression was added with https://github.com/vector-im/riot-android/pull/1900 where the `<img>` tag has been added without a check on its `src` attribute value.

PR #2056, opened 2018-03-07
Bug Fix: Embedded images must be restricted to mxc: URIs

giomfo:
#1921


manuroe:
Images with non expected URL appear like this:

Issue #2051, opened 2018-03-07
Privacy: Riot should never pass events content to GCM


manuroe:
at least without the clear consent from the end user.

When the [background sync](https://github.com/vector-im/riot-android/blob/e0e703d09a8b330d2779e75485c0ad0181d7452a/docs/notifications.md#background-sync-disabled) setting is disabled, Riot asks Sygnal to send matrix events content to GCM so that Riot can directly display detailed notifications from the GCM notification payload.

This is efficient for the battery and the network usage but this is a privacy concern. 
This is even a privacy issue because the app does not inform end users about that (see #2044) and what is behind this setting.

There are 2 approaches to fix this:
- never send content to GCM
When the setting is disabled, the app must not ask Sygnal to send event content to GCM. The user will see less detailed notifications like `X new messages in RoomY`.
- add another setting when *background sync* is disabled: `"Keep full detailed notifications"` with a subtitle `"This requires to send non encrypted message content through Google services servers"`.

ylecollen:
@manuroe 
1- "Riot should never pass events content to GCM"  seems wrong : Riot should never request the event contents when background sync is disabled

2 - it might also be fixed server side : many HS owners updated the GCM management to keep only the eventID/RoomId/nbr of notified rooms

3- to be sure that your privacy is kept, you should never used unencrypted rooms. Else the HS administrators could dispatch what you said.
The GCM payload does not have the message body if the room is encrypted.

4- as Master Coffee said, Riot will need to have an "advanced settings" or so. 

**Disable background sync** was implemented to save battery because the sync responses were too verbosed with presence.
By now, it seems that many users would prefer having a **stop notifying me** feature.

An endless story

manuroe:
> 1- "Riot should never pass events content to GCM" seems wrong : Riot should never request the event contents when background sync is disabled

@ylecollen, could you please elaborate?
I cannot see what is the problem with the title. I (well, we ;)) fail to see the difference between these 2 sentences.
The isue we want to solve here is to avoid do send messages contents to GCM, at least without informing the user. 
Like in e2e messages, sending clear message metadata is fine, ie we do not need to inform the user.
But we need to warm them when an application setting make their messages contents go to GCM.

> 2 - it might also be fixed server side

I agree. HS can enforce this. But this is the role of the app to inform the user what it is sent to third parties like GCM.

> 3- to be sure that your privacy is kept, you should never used unencrypted rooms

[...]

PR #2059, opened 2018-03-08
Bug Fix Privacy: Riot should never pass events content to GCM

giomfo:
Add another setting when background sync is disabled: "Keep full detailed notifications" with a subtitle "This requires to send non encrypted message content through Google services servers".

#2051


giomfo:
@manuroe what do you think to move this new setting in the `Notifications` section?
(It would be hidden if background sync is enabled)

giomfo:
Updated settings display:
screen shot 2018-03-09 at 10 40 35

Tyuoli:
It may be necessary to inform the user of the interest of activating or not the option of background sync. 
For example, we could specify the interest of not depending on Google for notifications and / or the interest of saving the battery.

manuroe:
The notification is really poor. But I can understand that the android code is not ready to provide the display name of the room without loading everything from the store.

giomfo:
@manuroe @Tyuoli : please review my last commit in this PR. The aim is to improve the notifications when the background sync is disabled.
The room name is now displayed:
screen shot 2018-03-10 at 01 37 31

hrj:
@giomfo @manuroe Thanks for seeing this through :heart:

Issue #2081, opened 2018-03-15
Crash on event with unexpected timestamp


manuroe:
The app crashed in [#offtopic](https://matrix.to/#/!UcYsUzyxTGDxLBEvLz:matrix.org/$1521087446409wOIFZ:kek.community) because it does not like events like:

```
{
  "origin_server_ts": 7404744525578619000,
  "sender": "@walle303:kek.community",
  "event_id": "$15210618456neDAh:kek.community",
  "age": 2399913,
  "unsigned": {
    "age": 2399913
  },
  "content": {
    "body": "i told it to give all messages a randomized 64 bit timestamp",
    "msgtype": "m.text"
  },
  "room_id": "!UcYsUzyxTGDxLBEvLz:matrix.org",
  "user_id": "@walle303:kek.community",
  "type": "m.room.message"
}
```

manuroe:
Discussion at https://matrix.to/#/!GnEEPYXUhoaHbkFBNX:matrix.org/$152106114611NlDRK:kek.community.

manuroe:
I renamed the title because the term crash seems to have been overestimated. Bad event TS introduces bad UI like in other Riot clients.

However, there is a bug that exists only on Android where such messages stay at the bottom at the history like at https://matrix.to/#/!GnEEPYXUhoaHbkFBNX:matrix.org/$1521111998151hKJuj:msg-net.de.

PR #2083, opened 2018-03-15
BF: Room: Events with unexpected timestamps get stuck at the bottom of the history

manuroe:
#2081

Use the more recent Event.isUndeliverable() method rather than the hacky Event.isValidOriginServerTs() (which works well for the display BTW)


manuroe:
Proof that it works:
(the buggy message is `yep, riot-android needs proper timestamp handling`)

PR #2100, opened 2018-03-20
change default piwik setting for f-droid version

mvgorcum:
As discussed a couple of times in #riot-android:matrix.org, changing the default setting for piwik analytics to off for the f-droid version will remove the tracking anti-feature in the f-droid store. Matthew said that changing this to remove the antifeature flag would be ok.

I think I did this right, but if I'm wrong let me know.

Signed-off-by: Mathijs van Gorcum <mvgorcum@gmail.com>


bmarty:
Hi @mvgorcum
Thanks for the PR.
You have duplicated an entire class to make just one little change on a class, which is not efficient.
Moreover, the Settings screen will not display the default value to "On" at first launch
I reject your PR, and submit a new one: #2176 
Benoît

PR #2121, opened 2018-03-29
Display sticker description on row message click

Tyuoli:
https://github.com/vector-im/riot-meta/issues/146

### DIFF
--- a/vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
+++ b/vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
@@ -1318,7 +1318,6 @@ public class VectorMessagesAdapter extends AbstractMessagesAdapter {
                 imageTypeView.setVisibility(View.GONE);
             }
 
-
             if (null != message) {
                 mHelper.hideStickerDescription(convertView);
 
@@ -1876,8 +1875,9 @@ public class VectorMessagesAdapter extends AbstractMessagesAdapter {
             tsTextView.setVisibility(View.VISIBLE);
         }
 
-        StickerMessage stickerMessage = JsonUtils.toStickerMessage(event.getContent());
+        // Show the description of the sticker only on message row click
         if (Event.EVENT_TYPE_STICKER.equals(event.getType())) {
+            StickerMessage stickerMessage = JsonUtils.toStickerMessage(event.getContent());
             if (null != stickerMessage && isInSelectionMode && isSelected) {
                 mHelper.showStickerDescription(contentView, stickerMessage);
             }
--- a/vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
+++ b/vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
@@ -1032,7 +1032,7 @@ public class VectorMessagesAdapter extends AbstractMessagesAdapter {
 
         convertView.setClickable(true);
 
-        // click on the avatar opens the details page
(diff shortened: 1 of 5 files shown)


Tyuoli:
stickers3
stickers2
stickers1

vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
```diff
@@ -1873,6 +1876,13 @@ private void manageSelectionMode(final View contentView, final Event event, fina
             tsTextView.setVisibility(View.VISIBLE);
         }
 
+        StickerMessage stickerMessage = JsonUtils.toStickerMessage(event.getContent());
```
manuroe: Should be done after having checked Event.EVENT_TYPE_STICKER.

vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
```diff
@@ -1320,8 +1320,11 @@ private View getImageVideoView(int type, final int position, View convertView, V
 
 
             if (null != message) {
+                mHelper.hideStickerDescription(convertView);
```
manuroe: Could we hide it by default in the xml layout?

vector/src/main/res/drawable/sticker_description_background.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android" android:shape="rectangle">
+    <corners android:radius="10dp" />
+    <solid android:color="@color/list_divider_color_light" />
```
ylecollen: did you check with theme update ? ie black / dark theme ?

vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
ylecollen: this code is needed because of recycled views but you're right, it should be hidden by default.

vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
ylecollen: you have msgType

Tyuoli: It's not a msgType from server but a new one in local to hack json parsing. It was created in order to obtain a StickerMessage even if it's an event type. 
It allows us to have a StickerMessage that inherits from ImageMessage. However, you can not use the msgType to do the checks.

vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
Tyuoli: I agree to hide the description of the sticker by default in the xml. 
However, I have to keep the hideStickerDescription method when calling the image view, otherwise it remains displayed after the click. I therefore lose the behavior where the description appears when the message row is selected and disappears when it is deselected (via the booleans : isInSelectionMode & isSelected).

vector/src/main/res/drawable/sticker_description_background.xml
Tyuoli: Thanks to point that. I will check.

Tyuoli:
**Sticker display in dark theme**

sticker_dark_theme



**Sticker display in black theme**

sticker_black_theme

PR #2122, opened 2018-03-30
Adapt message menu to stickers

Tyuoli:
Keep only Redact, View Source & Permalink


Tyuoli:
**Sticker menu**

sticker_menu

Issue #2097, opened 2018-03-19
Render sticker events in the timeline 


manuroe:
vector-im/riot-meta#146

PR #2160, opened 2018-04-17
Display animated gif stickers with Ion library

Tyuoli:
https://github.com/vector-im/riot-android/issues/2097


vector/src/main/java/im/vector/adapters/VectorMessagesAdapterMediasHelper.java
```diff
@@ -374,6 +383,7 @@ public void onDownloadComplete(String aDownloadId) {
             downloadProgressLayout.setVisibility(View.GONE);
         }
 
+        imageView.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
```
manuroe: Is it required by ion or is it old code to make the webview work ;) ?

If it is required, it would be better to do it only for stickers to avoid performance decrease.

PR #2163, opened 2018-04-19
Use Glide library to display sticker

Tyuoli:
Update libraries version

https://github.com/vector-im/riot-android/issues/2097

### DIFF
--- a/vector/build.gradle
+++ b/vector/build.gradle
@@ -142,10 +142,10 @@ dependencies {
 
     compile 'me.leolin:ShortcutBadger:1.1.2@aar'
 
-    compile 'com.android.support:appcompat-v7:26.1.0'
-    compile 'com.android.support:design:26.1.0'
-    compile 'com.android.support:cardview-v7:26.1.0'
-    compile 'com.android.support:recyclerview-v7:26.1.0'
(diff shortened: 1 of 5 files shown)


vector/build.gradle
```diff
@@ -142,10 +142,10 @@ dependencies {
 
     compile 'me.leolin:ShortcutBadger:1.1.2@aar'
 
-    compile 'com.android.support:appcompat-v7:26.1.0'
-    compile 'com.android.support:design:26.1.0'
-    compile 'com.android.support:cardview-v7:26.1.0'
-    compile 'com.android.support:recyclerview-v7:26.1.0'
+    compile 'com.android.support:appcompat-v7:27.1.1'
+    compile 'com.android.support:design:27.1.1'
```
bmarty: compileSdkVersion is set to 26, you should not use v27 of the lib

vector/build.gradle
```diff
@@ -204,6 +208,7 @@ dependencies {
     appCompile 'com.google.firebase:firebase-messaging:11.8.0'
 
     // fdroid flavor only
+    compile project(path: ':matrix-sdk')
```
bmarty: Why this line has been added?

Issue #2153, opened 2018-04-13
Contact details - Hide "Show Device List" option for non-Riot users


giomfo:

bmarty:
I closed the issue

PR #2165, opened 2018-04-19
Hide "Show devices list" for local contacts who are not matrix users

Tyuoli:
https://github.com/vector-im/riot-android/issues/2153

### DIFF
--- a/vector/src/main/java/im/vector/activity/VectorMemberDetailsActivity.java
+++ b/vector/src/main/java/im/vector/activity/VectorMemberDetailsActivity.java
@@ -1080,7 +1080,8 @@ public class VectorMemberDetailsActivity extends MXCActionBarActivity implements
             mListViewAdapter.setCallActionsList(callActions);
 
             // devices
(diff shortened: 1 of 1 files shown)


vector/src/main/java/im/vector/activity/VectorMemberDetailsActivity.java
```diff
@@ -1080,7 +1080,8 @@ private void updateAdapterListViewItems() {
             mListViewAdapter.setCallActionsList(callActions);
 
             // devices
-            if (mUser != null) {
+            // don't show devices list if the member isn't a matrix user
+            if (mUser != null && MXSession.PATTERN_CONTAIN_MATRIX_USER_IDENTIFIER.matcher(mMemberId).matches()) {
```
giomfo: You just need to check here that mMemberId is not null

giomfo: excuse I'm wrong

Issue #819, opened 2016-12-30
Built-in emoji support


cyphar:
Currently, you need to have an emoji keyboard installed in order to send emoji. Currently the `riot-web` client is planning on adding GitHub (or Slack) style `:emoji:`. This could work for Android, but we could also just follow how Signal handles it and add a button for inserting an emoji within the app. Also, on my phone the "enter" key is replaced by the "emoji" key inside Signal and the same doesn't happen on Android -- is there a chance we can set the same message entry flags to get emoji to be more prominent?

ylecollen:
The android keyboard already has an emoji keyboard.

maxlinux2000:
...but as default the android keyboard don't have the emoji active, and the users, don't know (and don't want to know) how activate it.

So the best think to do is to have an emoticons/emoji/stickers picker, like whatsapp/telegram/skype... an so on.
there are a lot of library that can be used... have a look here for ejample:
https://github.com/kontalk/androidclient  (emojicon)

Matrixcoffee:
What version of Android is this? My girlfriend bought a new Android (v6) phone and it definitely has the emoji button on the default Android keyboard.

vurpo:
There is some flag (it appears to be `android:inputType="textShortMessage"`) that can be set for the text box that makes the enter key on the keyboard change into an emoji button. This should be a trivial change to make in Riot.

ylecollen:
** textShortMessage ** will replace the Return key by the Emoji button.
It is not the expected behaviour.

Matrixcoffee:
I actually meant that there is a separate emoji button on Android's default keyboard, to the left of the spacebar, as well as the Return key. This emoji button is present on the keyboard when we tested it with Riot. (The emoji button is *always* present.)

So I am still curious which versions of Android do not have it.

eyelash:
Samsung Galaxy Young 2, Android 4.4.2, there is no way to enter emojis:
![screenshot_2017-01-25-18-06-34](https://cloud.githubusercontent.com/assets/2668288/22300966/8a2f1ee2-e329-11e6-833c-c5b7dbe35f5a.png)

derWalter:
lineage 14.1, no emoji button for riot present.

ylecollen:
Press and Hold the Return button offers to display the emoji keyboard.
http://www.wikihow.com/Get-Emoji-on-Android

eyelash:
> Press and Hold the Return button offers to display the emoji keyboard.

[...]

PR #2187, opened 2018-04-25
Add builtin emoji support

Foo-Manroot:
Before this series of commits, the emojis shown on the device where vendor-specific and relied on the keyboard to be inserted. Now, there's an integrated popup where one can insert emojis. Also, the displayed emojis are changed, as shown on the image:
screenshot

The library that I used was [vannkitec's](https://github.com/vanniktech/Emoji), but some changes were needed to integrate it on this project, so [I made a couple of modifications](https://github.com/vanniktech/Emoji/commit/581ca50a6f8a3ed650b5a5ef279439d2fa6b3af7) for that purpose. Until the new build is released, those changes are only available on the snapshot.

The emojis used are the ones from emoji-one for personal reasons (those were the ones I liked the most); but it can be changed easily modifying the dependency.

For more discussion about the need for built-in emoji support, refer to #819 

I tried to follow the contribution guides; but I may have missed something. In that case, I'd appreciate some comments and I'll try to fix whatever problem there may be.

Signed-off-by: Foo-Manroot <miguel.garciamartin@edu.uah.es>


bmarty:
Hi @Foo-Manroot ,

Thanks for the PR. Contribution guide has been followed :) . We will discuss internally to decide if we can merge it. Do you know what can be the impacts on the other client (iOS/Web) when emojis are sent?

Benoît

Foo-Manroot:
This is just an aesthetic change, as the message is the same. It just differs on its presentation to the user. On riot-desktop there's already a builtin emoji set (incidentally, the same set as the one I used, from EmojiOne), so they're viewed exactly the same.

As per other clients, they'll still see the emojis exactly the same as they've until now. For example, this is a comparative of this app before this pull request (the latest version on F-Droid), this app after the PR and Riot-desktop:

Before PR | After PR | Desktop
--------------|------------|------------
normal | with_mod | desktop

(the background colour of the second image is black because I had the two apps installed, and I wanted to distinguish between them... :D)

As I say, on the rest of clients the emojis will appear as always has been.

progserega:
When users use android-riot app - they see emoji from android. 
They are often better then on a PC.
If we are remove this "feature" and all emoji will be such as PC - this is will bad for users.
Compare emoji from Telegram and desktop riot:
1_riot_web
1_tele

2_riot_web
2_tele

Default emoji in riot is ugly.
This is no problem fore geeks and hackers - they are use textual smiles such :-) and :), :( 
But it is major for any other girls and womans. We 

We do not want the matrix to use only geeks? I thin we want, that matrix will use many many people. :-)

Foo-Manroot:
I... don't really know what's your point, @progserega; but I think there has been a consensus over at #android:matrix.org about giving the option to use these new font or the default one through configuration.

Right now I'm quite busy and I don't think I may be able to take a look at it in a while, but I encourage anyone to try to implement this as configurable feature.

progserega:
ok

PR #2192, opened 2018-04-26
Bug Fix: E2E: App keeps displaying ignored keyshare requests

### LINKED ISSUE #2191: E2E: App keeps displaying ignored keyshare requests
If the user taps on the "Ignore request" button of the keyshare request dialog, the app will show again the same dialog for the same request at the next startup.

manuroe:
Closes #2191.
Needs https://github.com/matrix-org/matrix-android-sdk/pull/258.

### DIFF
--- a/vector/src/main/java/im/vector/KeyRequestHandler.java
+++ b/vector/src/main/java/im/vector/KeyRequestHandler.java
@@ -1,6 +1,7 @@
 /*
  * Copyright 2016 OpenMarket Ltd
  * Copyright 2017 Vector Creations Ltd
+ * Copyright 2018 New Vector Ltd 
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


vector/src/main/java/im/vector/KeyRequestHandler.java
```diff
@@ -358,7 +373,7 @@ public void onSuccess(Void info) {
         mAlertDialog.setOnCancelListener(new DialogInterface.OnCancelListener() {
             @Override
             public void onCancel(DialogInterface dialogInterface) {
-                onDisplayKeyShareDialogClose(false);
+                onDisplayKeyShareDialogClose(false, true);
```
bmarty: 2nd param should be false in this case, no? User does not want explicitly ignore the request.
Nevermind I think this code is useless cause the dialog is not cancellable

manuroe: Well, I did not know what setOnCancelListener does. Sorry :|
I let you update it if needed.

PR #2248, opened 2018-05-04
Fix regression on WidgetApi due to Retrofit 2 upgrade

bmarty:


vector/src/main/java/im/vector/widgets/WidgetsRestClient.java
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright 2015 OpenMarket Ltd
+ * Copyright 2018 New Vector
```
manuroe: New Vector Ltd

Issue #2172, opened 2018-04-21
Piwik connections on opt-out until restart


licaon-kter:
No matter if you [ask the user for analytics approval](https://github.com/vector-im/riot-android/issues/1798) or if it's [opted out by default in F-Droid](https://github.com/vector-im/riot-android/pull/2100)... could you make the app **stop** trying to connect to `piwik.riot.im` **if** the user toggles opt-out to ON in Settings?

Thanks

Riot.im 0.8.5 F-Droid

Matrixcoffee:
Fixed in #2176

Bubu:
This is about a different issue tough, right?

Matrixcoffee:
Uh, right. I think I need more coffee...

Bubu:
Looking at the code there shouldn't be any connections when the preference is set to disabled. @licaon-kter where did you see this connection still happening?

licaon-kter:
NetGuard

![pic](https://image.ibb.co/envgoc/Image824722464395134392.jpg)

Matrixcoffee:
Did you confirm the setting is actually correct?

bmarty:
PR #2100 was bugged and as been rejected, replaced with #2176 
I've check in the code and no piwik web request should be done if opt-out is on
#2176 has been merged in the new version 0.8.7 (not yet published), please retry on this new version
Benoît

licaon-kter:
Ok I've figured it out.

The setting does not get applied right away, the user believes that toggling the switch does it (and let's be honest, I started the app to use it not turn it on and off at random), BUT the app needs to be EXITed first and then it stops connecting.

Either [modify the text](https://github.com/vector-im/riot-android/blob/develop/vector/src/main/res/values/strings.xml#L613) to say `Riot collects anonymous analytics to allow us to improve the application. (Restart needed)` or force a restart on toggle.

zoff99:
opt-out itself is not privacy friendly. it should be opt-in in the first place

licaon-kter:
@zoff99 use F-Droid, problem solved

Bubu:
Pretty sure this can be closed now with the gdpr stuff present?

PR #2261, opened 2018-05-14
Piwik opt-out needs a restart

licaon-kter:
fix https://github.com/vector-im/riot-android/issues/2172


bmarty:
Thanks for the PR. I'm not sure it will be merged for two reasons:
- String has to be edited with the Weblate tool
- The real fix should be to disable piwik without the necessity to restart application

licaon-kter:
Workflow is usually github-weblate-translations to other that english, it is not like that here?

I agree, but in the meantime...

bmarty:
I think piwik opt-out does not need a restart anymore. I'll check soon to safely close this PR

bmarty:
Analytics has been improved, no restart is needed anymore

PR #2274, opened 2018-05-17
Registration bug fix

giomfo:
- Fix the error "Unable to register" (password missing), when the email is validated quickly.
- Remove the pending email validation params on registration failure.

related to #2150

### DIFF
--- a/vector/src/main/java/im/vector/RegistrationManager.java
+++ b/vector/src/main/java/im/vector/RegistrationManager.java
@@ -207,6 +207,8 @@ public class RegistrationManager {
                 @Override
                 public void onRegistrationSuccess() {
                     // The registration could not succeed without password.
+                    // Keep calling listener (the error case) as a fallback,
+                    listener.onUsernameAvailabilityChecked(false);
                 }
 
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/RegistrationManager.java
```diff
@@ -199,23 +199,19 @@ public void setSupportedRegistrationFlows(final RegistrationFlowResponse registr
...
+            // Trigger a fake registration (without password) to know whether the user name is available or not.
+            RegistrationParams params = new RegistrationParams();
             params.username = mUsername;
-            params.password = mPassword;
 
             register(context, params, new InternalRegistrationListener() {
                 @Override
                 public void onRegistrationSuccess() {
-                    listener.onUsernameAvailabilityChecked(true);
+                    // The registration could not succeed without password.
```
bmarty: It's quite dangerous to remove the callback in this case (trying to avoid infinite loading view)
You should keep calling `listener.onUsernameAvailabilityChecked(false);` (the error case) as a fallback, for symetry

PR #2282, opened 2018-05-21
remove # before build version

krombel:
Every time when people refer to a riot-build they use the string
as provided in the app. This leads to a trigger of the github bot
which thinks that the version string refers to a issue/ pull request

To avoid this we replace the hashtag with a b so we now get a version
like G-b1234. This is currently not used anywhere as trigger (AFAIK)
so the room should not be polluted with content of overactive bots

Signed-Off-by: Matthias Kesler <krombel@krombel.de>

### DIFF
--- a/vector/src/main/java/im/vector/Matrix.java
+++ b/vector/src/main/java/im/vector/Matrix.java
@@ -235,7 +235,7 @@ public class Matrix {
         String buildNumber = mAppContext.getResources().getString(R.string.build_number);
 
         if ((useBuildNumber) && !TextUtils.equals(buildNumber, "0")) {
-            gitVersion = "#" + buildNumber;
+            gitVersion = "b" + buildNumber;
             longformat = false;
         }
 


bmarty:
Thanks for the PR.
I understand the issue, nevertheless, "b" can be a digit for a commit id. So it collapses with the version number used in prod which uses commit id (ex: `G-b0f291d0`). I think there no pb if you change "b" to "build ". Let's be verbose :)
@giomfo @manuroe ok for you?
Benoit

giomfo:
@bmarty If I follow correctly, the change will impact only the builds from jenkins, so the version here is a number (the change does not concern app version based on commit id).
So feel free to use "b" or "build "
I would vote for "build " ;)

krombel:
@bmarty As @giomfo said: Only relevant for build which have a jenkins build id - so that should always be numeric
I would be OK with every character besides "#". I can update the PR with another character if you have preferences for another one.
And concerning "b" vs "build": I would vote for "b" as "build" might be too long in some situations.

bmarty:
Ok for 'b'. Build number is shorter than commit Id

Issue #2284, opened 2018-05-22
Settings: Add deactivate account 


manuroe:
with a checkbox for GDPR erasure.

https://github.com/vector-im/riot-meta/issues/164

bmarty:
Done partially: it inform the user to do it on the web client

bmarty:
Yes indeed, thanks

PR #2293, opened 2018-05-25
Deactivate Account (redirect to web client)

bmarty:
Partially implements #2284: it only display a Screen redirecting to the web client

### DIFF
--- a/vector/src/main/java/im/vector/activity/CommonActivityUtils.java
+++ b/vector/src/main/java/im/vector/activity/CommonActivityUtils.java
@@ -341,7 +341,7 @@ public class CommonActivityUtils {
      * @param activity the caller activity
      */
     public static void logout(Activity activity) {
-        logout(activity, true);
+        logout(activity, true, true);
     }
 
     /**
@@ -350,7 +350,7 @@ public class CommonActivityUtils {
      * @param activity      the caller activity
      * @param goToLoginPage true to jump to the login page
      */
-    public static void logout(final Activity activity, final boolean goToLoginPage) {
+    public static void logout(final Activity activity, final boolean goToLoginPage, boolean clearSession) {
         Log.d(LOG_TAG, "## logout() : from " + activity + " goToLoginPage " + goToLoginPage);
 
         // if no activity is provided, use the application context instead.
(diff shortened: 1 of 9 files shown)


vector/src/main/java/im/vector/activity/CommonActivityUtils.java
```diff
@@ -350,7 +350,7 @@ public static void logout(Activity activity) {
      * @param activity      the caller activity
      * @param goToLoginPage true to jump to the login page
      */
-    public static void logout(final Activity activity, final boolean goToLoginPage) {
+    public static void logout(final Activity activity, final boolean goToLoginPage, boolean clearSession) {
```
giomfo: Could you please add a comment about `clearSession`param

vector/src/main/java/im/vector/activity/CommonActivityUtils.java
```diff
@@ -379,51 +379,62 @@ public static void logout(final Activity activity, final boolean goToLoginPage)
...
+
+            // clear credentials
+            Matrix.getInstance(context).clearSessions(context, true, new SimpleApiCallback<Void>() {
+                @Override
+                public void onSuccess(Void info) {
+                    clearData(context, goToLoginPage);
+                }
+            });
+        } else {
+            clearData(context, goToLoginPage);
```
giomfo: Note "removeCredentials" should be required in this case too

vector/src/main/java/im/vector/activity/CommonActivityUtils.java
```diff
@@ -379,51 +379,62 @@ public static void logout(final Activity activity, final boolean goToLoginPage)
...
-                        activeActivity.startActivity(new Intent(activeActivity, LoginActivity.class));
-                        activeActivity.finish();
-                    } else {
-                        Intent intent = new Intent(context, LoginActivity.class);
-                        intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
-                        context.startActivity(intent);
-                    }
-                }
+        if (goToLoginPage) {
+            Matrix.getInstance(context).getSharedGCMRegistrationManager().clearPreferences();
```
giomfo: I think you already clearPreferences before see line 385

PR #2308, opened 2018-05-29
update the Simplified translation

tleydxdy:
according to the new english version and Tranditional Chinese version 
%s parts may need revising, the use of quotes seems inconsistant in both versions


tleydxdy:
oof, saw the italian one, will try to update via weblate.

bmarty:
Thanks @tleydxdy :)

PR #2311, opened 2018-05-29
Send sticker: open integration manager from the sticker picker

bmarty:
- Enable the feature
- Open the integration manager if the widget is install but no sticker pack is available, when the user clicks on the link


vector/src/main/java/im/vector/activity/AbstractWidgetActivity.kt
```diff
@@ -236,7 +242,35 @@ abstract class AbstractWidgetActivity : RiotAppCompatActivity() {
...
+
+        when (action) {
+            "integration_manager_open" -> {
+                var integType: String? = null
+                var integId: String? = null
+
+                val data = eventData["data"]
+
+                data?.let {
+                    val dict = data as JsonDict<String>
```
manuroe: It may worth to check class type in `data` values.

bmarty: Yes, I've added the checks

PR #2322, opened 2018-05-30
Account deactivation on Riot Android

bmarty:
Account deactivation is now managed on the Android application

This PR also include:
- build script enhancement (along with Jenkins configuration change)

- creation of a DebugReceiver, to help debugging the application. It is only activated on DEBUG versions. Current supported actions are:
    - `ls` on the application filesystem
    - dump of the SharedPref.


bmarty:
Note: https://github.com/matrix-org/matrix-android-sdk/pull/275 is needed, so Travis will fail :(

vector/src/main/java/im/vector/activity/CommonActivityUtils.java
```diff
@@ -427,6 +428,77 @@ public void onSuccess(Void info) {
...
+
+                try {
+                    ShortcutBadger.setBadge(context, 0);
+                } catch (Exception e) {
+                    Log.d(LOG_TAG, "## logout(): Exception Msg=" + e.getMessage());
+                }
+
+                // warn that the user logs out
+                Collection<MXSession> sessions = Matrix.getMXSessions(context);
+                for (MXSession session : sessions) {
```
manuroe: We could make that only for `mxSession`

bmarty: ok, yes, done

PR #2339, opened 2018-06-05
Another PR with code cleanup

bmarty:
- Reduce the size of CommonActivityUtils and VectorApp by creating dedicated classes
- remove unnecessary call to getResource()
- rework the Theme manager and convert to Kotlin
- move some methods and convert them to Kotlin extension
- rework menu management (reduce duplication of code)
- remove duplicated file google-services.json
- rename UTILS_OPACITY_NONE to UTILS_OPACITY_FULL because the value is 1.0 and UTILS_OPACITY_FULL to UTILS_OPACITY_NONE because the value is 0.0


vector/src/main/java/im/vector/util/ThemeUtils.kt
```diff
@@ -0,0 +1,214 @@
+/*
+ * Copyright 2017 OpenMarket Ltd
```
giomfo: You should set here the 2018 copyright here

vector/src/main/java/im/vector/util/ThemeUtils.kt
```diff
@@ -0,0 +1,214 @@
...
+    /**
+     * Update the menu icons colors
+     *
+     * @param menu  the menu
+     * @param color the color
+     */
+    fun tintMenuIcons(menu: Menu, color: Int) {
+        for (i in 0 until menu.size()) {
+            val item = menu.getItem(i)
+            val drawable = item.icon
```
giomfo: Can we consider here a potential item.title?
(see https://github.com/dinsic-pim/riot-android/pull/172/files#diff-d2c1df037c6ac766c6370a262f602269R2129)

bmarty: As discussed with @Tyuoli , it is ok like that

vector/src/main/java/im/vector/util/ThemeUtils.kt
bmarty: done

PR #2342, opened 2018-06-06
Check too long files - limit set to 3000 for the moment

bmarty:


bmarty:
We can see that Travis Fail cause 2 files has more than 3000 lines :)

PR #2343, opened 2018-06-06
Rework Activity menus and DialogBuilder

bmarty:
Code quality, no functional impact

### DIFF
--- a/vector/src/main/java/im/vector/activity/VectorHomeActivity.java
+++ b/vector/src/main/java/im/vector/activity/VectorHomeActivity.java
@@ -545,14 +545,14 @@ public class VectorHomeActivity extends RiotAppCompatActivity implements SearchV
             // crash reported by a rage shake
             try {
                 new AlertDialog.Builder(this)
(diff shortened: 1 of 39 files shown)


vector/src/main/java/im/vector/activity/VectorHomeActivity.java
```diff
@@ -1159,12 +1155,8 @@ private void displayCryptoCorruption() {
                 editor.putBoolean(isFirstCryptoAlertKey, false);
                 editor.commit();
 
-                AlertDialog.Builder alertDialogBuilder = new AlertDialog.Builder(this);
-                alertDialogBuilder.setMessage(getString(R.string.e2e_need_log_in_again));
-
-                // set dialog message
-                alertDialogBuilder
-                        .setCancelable(true)
```
manuroe: `.setCancelable(true)` has disappeared. Is it expected?

bmarty: @manuroe yes, AlertDialogs are cancellable by default, so this call was not necessary

PR #2362, opened 2018-06-12
use apply() instead of commit() to save shared prefs

### LINKED ISSUE #2231: Improve SharedPreferences usage and implementation
- Keys should be centralized in a separate file to prevent collapsing

- In many cases, `preferences` and `editor` can be inlined

- Editor instance can be mutualized in `PreferenceManager.fixMigrationIssues()`

- `SharedPreferences.Editor.apply()` should be used instead of `SharedPreferences.Editor.commit()`

bmarty:
Fixes #2231

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ Features:
 
 Improvements:
  - Add spacing to device keys (#2314)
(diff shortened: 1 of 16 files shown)


bmarty:
Thx

PR #2363, opened 2018-06-13
Improve code quality for notification and fix #2354

### LINKED ISSUE #2354: Join button doesn't work in the invite notification
screenshot_riot im_20180608-225052

bmarty:
Convert some classes to Kotlin

Fixes #2354


vector/src/main/java/im/vector/notifications/NotificationUtils.kt
```diff
@@ -0,0 +1,845 @@
...
+
+            } else {
+                resizedBitmap = Bitmap.createBitmap(
+                        bitmap,
+                        0,
+                        (height - width) / 2,
+                        width,
+                        width
+                )
+            }// higher than large
```
manuroe: Comments are now away from the code they explain

bmarty: Ok, fixed. Also this code has been moved to a dedicated class, so we can remove some code duplication :)

Issue #2312, opened 2018-05-29
URL previews don't wrap


lampholder:
Reported alongside https://github.com/vector-im/riot-web/issues/6693

bmarty:
It's better now:

![urlpreviewbetter](https://user-images.githubusercontent.com/3940906/42207684-ce54e4b2-7eaa-11e8-9f01-2fb553bfc0fe.png)

bmarty:
Replaced by #2574

PR #2405, opened 2018-07-03
Feature/fix issue before release

bmarty:
Fix issues in UrlPreviews (#2312)
Fix some other issues on opening external applications


vector/src/main/java/im/vector/util/SlashCommandsParser.java
```diff
@@ -95,6 +95,9 @@
     // on / off
     private static final String CMD_MARKDOWN = "/markdown";
 
+    // clear scalar token (waiting for correct 403 management)
+    private static final String CMD_CLEAR_SCALAR_TOKEN = "/clear_scalar_token";
```
manuroe: Is it a command available on riot-web?

bmarty: I don't think so

manuroe: as long there is an issue in riot-meta for that to track parity across Riot clients, I am fine.

bmarty: https://github.com/vector-im/riot-meta/issues/208

PR #2423, opened 2018-07-11
add silent counting notification in case event is not parseable

krombel:
In the case the event cannot be parsed show now the unreadMessageCount instead of no notification at all


bmarty:
Hi @krombel ,

Thanks for this PR. How do you reproduce the case where an event cannot be parsed?

Benoît

PR #2424, opened 2018-07-11
allow overwriting DEFAULT_PUSHER_URL via Settings store

krombel:
To experiment with push services I found it useful to overwrite the
push gateway directly on the device. For this the already integrated
"Stheto" framework is used.

A future plan might be to implement some UI for that but for
DEV-purposes that is enough.

Signed-Off-By: Matthias Kesler <krombel@krombel.de>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,7 @@ Changes in Riot 0.8.13 (2018-XX-XX)
 ===================================================
 
 Features:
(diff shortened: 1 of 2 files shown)


krombel:
This currently makes (in case of GCM) only sense if you also find a way to adapt the keys which come from [google-services.json](https://github.com/vector-im/riot-android/blob/develop/vector/src/app/google-services.json).
I am going to close this for now as changing that values would need a recompilation either way.

Issue #2417, opened 2018-07-07
Phone call accept button missing


chris-je:
Hey,
Today I updated my Matrix client to version 0.8.12 (G-b146) on my HTC Android 6.0.1.
Since this version, i am not able to accept incoming phone calls.
The accept button doesn't exist anymore, there is just the a hang up button.

bmarty:
To investigate

bmarty:
(thanks @chris-je for reporting this issue)

bmarty:
@chris-je on which screen you don't see the accept button anymore?
- The Room list (home)?
- in a Room?
- the notification?
- Somewhere else?

Thanks for this precision

Luraktinus:
i can confirm, here a Screenshot of the "Incoming Call" screen
@bmarty 
![screenshot_20180713-102507](https://user-images.githubusercontent.com/33608674/42681384-d06d9e94-8687-11e8-8320-3c3b3c192896.png)

bmarty:
Reproduced when the phone is in "Do not disturb" mode

dva-re:
dev build 2258, bug still there regardless of "Do not disturb" mode @bmarty 
![image](https://user-images.githubusercontent.com/10573161/42819469-54c1e45a-89dc-11e8-82aa-9f1c43342434.png)

Luraktinus:
can't reproduce :(

dva-re:
I was send rageshake report with comment 'for issue 2417'

bmarty:
https://github.com/matrix-org/riot-android-rageshakes/issues/2858

digititus:
Confirmed on build #2278. Not able to answer voice or video calls

dva-re:
It starts from build 2247

dva-re:
fixed

PR #2428, opened 2018-07-13
Fix issue on incoming call screen when "Do not disturb mode" is active (#2417)

bmarty:
Request https://github.com/matrix-org/matrix-android-sdk/pull/317


vector/src/main/java/im/vector/util/CallsManager.java
```diff
@@ -538,6 +538,8 @@ public void onHangUp(String hangUpMsg) {
     private void startRinging() {
         if (NotificationUtils.INSTANCE.isDoNotDisturbModeOn(mContext)) {
             Log.w(LOG_TAG, "Do not ring because DO NOT DISTURB MODE is on");
+            mCallSoundsManager.startRingingSilently();
+
```
ganfra: Why the extra line break here?

Issue #394, opened 2016-08-18
Send Text With Enter Key


cwmke:
It would great to have an option to have the enter/send key submit the text instead of creating a new line. I know many other chat/sms apps offer this as an option in settings. Thanks for any consideration.

sbts:
This is definitely a duplicate of #220 can someone mark it as such please

PR #2458, opened 2018-07-24
Send Text With Enter Key

Tyuoli:
Related to #394


vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -2313,6 +2317,40 @@ public void onRequestPermissionsResult(int aRequestCode, @NonNull String[] aPerm
...
     }
 
+    /**
+     * The user can use enter key on his soft keyboard to add a new line or to send message
+     * depending on the settings he has chosen.
+     */
+    private void manageKeyboardOptionsToSendMessage() {
+        if (PreferencesManager.useEnterKeyToSendMessage(this)) {
+            mEditText.setImeOptions(EditorInfo.IME_ACTION_DONE);
+            mEditText.setOnKeyListener(new View.OnKeyListener() {
```
ganfra: You should use setOnEditorActionListener instead

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -721,6 +722,8 @@ public boolean onEditorAction(TextView textView, int actionId, KeyEvent keyEvent
             }
         });
 
+        manageKeyboardOptionsToSendMessage();
```
ganfra: Why having this in onCreate and onResume ?

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -2313,6 +2317,40 @@ public void onRequestPermissionsResult(int aRequestCode, @NonNull String[] aPerm
...
+                        }
+                    }
+                    return false;
+                }
+            });
+        } else {
+            mEditText.setImeOptions(EditorInfo.IME_ACTION_UNSPECIFIED);
+            mEditText.setInputType(InputType.TYPE_TEXT_FLAG_MULTI_LINE);
+            mEditText.setOnKeyListener(null);
+            mEditText.setSingleLine(false);
```
ganfra: Question, is multiline not compatible with send action ?

Tyuoli: I will test

Tyuoli: Unfortunately no.

vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -2323,6 +2325,34 @@ public void onRequestPermissionsResult(int aRequestCode, @NonNull String[] aPerm
...
+                    if (actionId  == EditorInfo.IME_ACTION_SEND) {
+                        sendTextMessage();
+                        return true;
+                    }
+                    return false;
+                }
+            });
+        } else {
+            mEditText.setImeOptions(EditorInfo.IME_ACTION_UNSPECIFIED);
+            mEditText.setInputType(InputType.TYPE_TEXT_FLAG_MULTI_LINE);
```
bmarty: `textCapSentences` is lost then. It should not

Issue #2430, opened 2018-07-13
Flair section title in the setting is displayed even if user has no flair


bmarty:
![flair](https://user-images.githubusercontent.com/3940906/42690237-c2982af2-86a3-11e8-81c9-844e7576f3e6.png)

manuroe:
We should do like web and displays "You're not currently a member of any communities." to invite users to use the feature. There is a screenshot or riot-web at https://github.com/vector-im/riot-ios/issues/1948.

PR #2460, opened 2018-07-24
Flair section title in the setting is displayed even if user has no...

Tyuoli:
…flair

Related to #2430

### DIFF
--- a/vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
+++ b/vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
@@ -189,9 +189,6 @@ class VectorSettingsPreferencesFragment : PreferenceFragment(), SharedPreference
     private val mLabsCategory by lazy {
         findPreference(PreferencesManager.SETTINGS_LABS_PREFERENCE_KEY) as PreferenceCategory
     }
-    private val mGroupsFlairCategory by lazy {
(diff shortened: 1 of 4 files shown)


Tyuoli:
To fix this issue, we have 2 options:
1) Keep the flair section and add the title only if the user has flairs
2) Delete the flair section and add it only if the user has flairs

I do not know what is the best but here is the display for the 2 options: 

**Before:**
capture d ecran 2018-07-26 a 10 04 20

**Option 1: without flair vs with flair**

capture d ecran 2018-07-26 a 11 32 36

capture d ecran 2018-07-26 a 11 34 33

**Option 2: without vs with flair**

capture d ecran 2018-07-26 a 09 55 07

capture d ecran 2018-07-26 a 09 52 41

Tyuoli:
Result : 
capture d ecran 2018-07-31 a 11 34 00

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
```diff
@@ -2589,15 +2591,15 @@ class VectorSettingsPreferencesFragment : PreferenceFragment(), SharedPreference
      * It can be any mobile device, as any browser.
      */
     private fun refreshGroupFlairsList() {
-        // display a spinner while refreshing
-        if (0 == mGroupsFlairCategory.preferenceCount) {
-            val preference = ProgressBarPreference(activity)
-            mGroupsFlairCategory.addPreference(preference)
```
bmarty: Why removing this code?

Issue #2384, opened 2018-06-25
Auto completion for / command


bmarty:
When user types /, it should be convenient to display the list of available commands in a popin, the same way we display completion for user name

manuroe:
@bmarty, this is a duplicate of https://github.com/vector-im/riot-android/issues/247, no?

This is a long waiting feature...

bmarty:
Sorry, haven't checked for any existing issue

bmarty:
Audrey, it's maybe not long to implement, re-using the popup we already have to suggest users

PR #2463, opened 2018-07-27
Auto completion for / command #2384

Tyuoli:
slash_command_auto_completion

 - TODO : Handle the i18n for slash command description

### DIFF
--- a/vector/src/main/java/im/vector/activity/VectorRoomActivity.java
+++ b/vector/src/main/java/im/vector/activity/VectorRoomActivity.java
@@ -1084,6 +1084,7 @@ public class VectorRoomActivity extends MXCActionBarActivity implements
         ViewedRoomTracker.getInstance().setViewedRoomId(null);
         ViewedRoomTracker.getInstance().setMatrixId(null);
         mEditText.initAutoCompletion(mSession, null);
+        //mEditText.initAutoCompletionCommandLine(mSession, null);
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -721,12 +721,26 @@ public boolean onEditorAction(TextView textView, int actionId, KeyEvent keyEvent
...
+        mEditText.initAutoCompletion(mSession, (null != mRoom) ? mRoom.getRoomId() : null);
+        mEditText.initAutoCompletionCommandLine(mSession , null);
         mEditText.setAddColonOnFirstItem(true);
 
         mEditText.addTextChangedListener(new TextWatcher() {
             @Override
             public void afterTextChanged(android.text.Editable s) {
                 if (null != mRoom) {
+                    if (mEditText.getText().toString().startsWith("@")) {
+                        mEditText.setAdapter(mEditText.mAdapterUser);
```
ganfra: You should maybe add an enum for type and a method to change it like :
setMode(AutoCompletion.Command) and setMode(AutoCompletion.User)
The method will be responsible of setting adapter and threshold when the mode is different from the one already set.

vector/src/main/java/im/vector/util/SlashCommandsParser.java
```diff
@@ -44,39 +44,63 @@
...
-    // <user-id> [<power-level>]
-    private static final String CMD_SET_USER_POWER_LEVEL = "/op";
-
-    // <user-id>
-    private static final String CMD_RESET_USER_POWER_LEVEL = "/deop";
+    public enum SlashCommand {
+
+        // defines the command line operations
+        // the user can write theses messages to perform some room events
+        EMOTE("/me", "<message>", VectorApp.getInstance().getString(R.string.command_description_emote)),
```
ganfra: You should let the enum have a ResInt instead String :
EMOTE("/me", "<message>", R.string.command_description_emote)

bmarty:
@Tyuoli please remove this branch if it is ok (this PR is replaced by #2476)

Issue #530, opened 2016-09-20
Provide a smarter notification icon


ylecollen:
![screenshot_2016-09-20-09-13-14](https://cloud.githubusercontent.com/assets/8971031/18660790/b9ad4e9e-7f12-11e6-9761-c2864a95168f.png)

giomfo:
Matthew, the Android users complain about the current notification icon.
Could you please check whether a smarter notification icon may be provided?

bmarty:
Hello @dkanada ,

Thanks for the proposal. Yes, feel free to submit a PR to update the notification icon. Please follow the Android guidelines here: https://developer.android.com/guide/practices/ui_guidelines/icon_design_status_bar (no need to take care of Android version before 3.0)

Thanks

Benoît

bmarty:
Done

PR #2492, opened 2018-08-06
modify notification icon

dkanada:
Mentioned in #530 about adding a solid icon for notifications on mobile. I took a vector of the logo and removed the inner three lines, then merged all the layers into one solid shape. The padding should be fine but let me know if it looks awkward, I used the normal amount for android icons.


dkanada:
One thing to note, the link you used above mentioned a small amount of padding for the icons, and the material design guidelines mention padding as well for app icons, but there wasn't any on the old notification icon. On 24 pixel rectangular shapes the suggested amount was two pixels on each side of the rectangle.

bmarty:
Hi @dkanada ,

Thanks for the PR. The 'R' is good, but we cannot see the 'I' included in the logo anymore. Isn't it possible to keep the 'I' visible? Maybe with a wired 'R'?

Also, could you please sign-off it as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off, and add a line in CHANGES.rst

Benoît

dkanada:
Here are three different variations with the "i" included. I will order them by what I think looks best first, I don't have a background in design but I think the solid character or the first one directly below looks best. If you can think of improvements let me know, I wasn't able to figure out a good way to incorporate the "i" without making it look a bit odd.

first
second
third

bmarty:
Thanks for the 3 proposals. I'll share this with the team tomorrow, and we will decide what we can do
Benoit

bmarty:
Hi @dkanada ,

The first proposal is good 👍 

Please don't forget my 2 other requests (CHANGES.rst and Sign-Off)

Benoît

dkanada:
Before I finish up this pull request, I made a few more small changes just to see what you thought. I noticed the navigation drawer was also using the same pixelated icon, so I changed it to use this new icon. There were three different copies of the same icon at that point (permanent notification, normal notification, and transparent logo) so I just renamed all references to transparent logo since that is a good name for this icon.

The messages and settings icons in the navigation drawer have always been slightly larger than the other three, and since I fixed the messages icon with padding I went ahead and fixed the settings icon as well. I used the default android settings icon which I grabbed from [here](material.io) which I think looks fine.

Last item, the new transparent logo looks fine in the navigation drawer but looks a bit odd in the notification bar. I thought of one more [possible design](https://user-images.githubusercontent.com/21353219/43858356-c60d9a38-9b88-11e8-9479-a3d55b4e8ef6.png) that might work.

dkanada:
It seems I accidently used the develop branch instead of master, let me know what you want me to do about that.

bmarty:
Hello @dkanada ,

- We validate the "Messages" icon
- We validate the proposal for the notification icon (https://user-images.githubusercontent.com/21353219/43858356-c60d9a38-9b88-11e8-9479-a3d55b4e8ef6.png)
- We validate the setting icon

Thanks!

dkanada:
Signed-off-by: Dillon Kanada <dkanada@protonmail.com>

dkanada:
Just removed the merge commit and rebased on develop, everything looks good now.

bmarty:
Thaks for update. Finally you've taken the last icon for the message menu as well, but it's ok, the entry should be removed in a near future, as it is totally useless...
I merge the PR

bmarty:
@dkanada people like the icon you've created 👍 
https://matrix.to/#/!XIydWLOEJvcDBKwWKE:matrix.org/$153546416554822cZcXf:matrix.org

Issue #2437, opened 2018-07-17
When user bans or un-bans someone, notify that it will restart the app


bmarty:
The first time I did that, I thougt it was a crash :)

Especially, as first sync can be quite long, If user is in low network coverage, he will be warned and may postpone his mute/unmute action

User can be muted in the user detail screen and un-muted in the settings

Tyuoli:
<img width="673" alt="capture d ecran 2018-07-24 a 11 52 41" src="https://user-images.githubusercontent.com/36439413/43130970-333f8cea-8f38-11e8-9e2e-5375b247ae27.png">

krombel:
This is especially an issue when you ban/unban sb. on e.g. riot-web. Then the app restarts without you getting an info on why.

PR #2493, opened 2018-08-07
When user bans or un-bans someone, notify that it will restart the app

Tyuoli:
Related to #2437 

capture d ecran 2018-08-06 a 17 37 37


bmarty:
(note: replaced by #2499)

Issue #2491, opened 2018-08-06
Improve GDPR consent webview management


manuroe:
https://github.com/vector-im/riot-meta/issues/217

PR #2498, opened 2018-08-07
Feature/better rgpd

ganfra:
#2491

### DIFF
--- a/vector/src/main/java/im/vector/activity/VectorWebViewActivity.kt
+++ b/vector/src/main/java/im/vector/activity/VectorWebViewActivity.kt
@@ -22,11 +22,18 @@ import android.os.Build
 import android.support.annotation.StringRes
 import android.webkit.WebChromeClient
 import android.webkit.WebView
-import android.webkit.WebViewClient
 import butterknife.BindView
 import im.vector.R
+import im.vector.webview.VectorWebViewClient
+import im.vector.webview.WebViewMode
 
-class SimpleWebViewActivity : RiotAppCompatActivity() {
+/**
+ * This class is responsible for managing a WebView
+ * It does also have a loading view and a toolbar
+ * It relies on the VectorWebViewClient
+ * This class shouldn't be extended. To add new behaviors, you might create a new WebViewMode and a new WebViewEventListener
+ */
+class VectorWebViewActivity : RiotAppCompatActivity() {
 
     /* ==========================================================================================
      * UI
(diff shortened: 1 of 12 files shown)


vector/src/main/java/im/vector/activity/VectorWebViewActivity.kt
```diff
@@ -35,35 +42,37 @@ class SimpleWebViewActivity : RiotAppCompatActivity() {
...
 
-    override fun getLayoutRes() = R.layout.activity_simple_web_view
+    override fun getLayoutRes() = R.layout.activity_vector_web_view
 
     override fun initUiAndData() {
         configureToolbar()
+        waitingView = findViewById(R.id.simple_webview_loader)
 
-        webView.settings.let {
+        webView.settings.apply {
```
bmarty: Ok to change from `let` to `apply`, but I had bad experiment when calling a method which exists both in the object and in the current class. There is no call to any method in the apply block, so ok for this change.

vector/src/main/java/im/vector/util/WeakReferenceDelegate.kt
```diff
@@ -0,0 +1,21 @@
+package im.vector.util
```
bmarty: Please add copyright

vector/src/main/java/im/vector/webview/ConsentWebViewEventListener.kt
```diff
@@ -0,0 +1,83 @@
...
+        if (TextUtils.equals(url, SUCCESS_URL)) {
+            createRiotBotRoom()
+        }
+    }
+
+    /**
+     * This methods try to create the RiotBot room when the user agreed
+     */
+    private fun createRiotBotRoom() {
+        safeActivity?.showWaitingView()
```
bmarty: Put this in the `let` block?

vector/src/main/java/im/vector/webview/ConsentWebViewEventListener.kt
```diff
@@ -0,0 +1,83 @@
...
+     * APICallback instance
+     */
+    private val createRiotBotRoomCallback = object : ApiCallback<String> {
+        override fun onSuccess(info: String) {
+            Log.d(LOG_TAG, "## onRegistrationSuccess() : succeed to invite riot-bot")
+            safeActivity?.finish()
+        }
+
+        private fun onError(error: String?) {
+            Log.e(LOG_TAG, "## onRegistrationError() : failed  to invite riot-bot $error")
```
bmarty: Maybe also finish the Activity in case of error?

vector/src/main/java/im/vector/webview/ConsentWebViewEventListener.kt
```diff
@@ -0,0 +1,83 @@
...
+            Matrix.getInstance(it).defaultSession.createDirectMessageRoom(RIOT_BOT_ID, createRiotBotRoomCallback)
+        }
+    }
+
+    /**
+     * APICallback instance
+     */
+    private val createRiotBotRoomCallback = object : ApiCallback<String> {
+        override fun onSuccess(info: String) {
+            Log.d(LOG_TAG, "## onRegistrationSuccess() : succeed to invite riot-bot")
```
bmarty: why "onRegistrationSuccess"?

vector/src/main/java/im/vector/webview/ConsentWebViewEventListener.kt
bmarty: Why "onRegistrationError"?

vector/src/main/java/im/vector/webview/DefaultWebViewEventListener.kt
```diff
@@ -0,0 +1,50 @@
...
+    override fun onPageStarted(url: String) {
+        Log.d(TAG, "On page started: $url")
+    }
+
+    override fun onPageFinished(url: String) {
+        Log.d(TAG, "On page finished: $url")
+    }
+
+    override fun onPageError(url: String, errorCode: Int, description: String) {
+        Log.v(TAG, "On received error: $url - error: $description")
```
bmarty: maybe also log errorCode since this class do log everything. And maybe use Log.e()

vector/src/main/java/im/vector/webview/DefaultWebViewEventListener.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.matrix.androidsdk.util.Log
+
+private const val TAG = "DefaultWebViewEventListener"
+
+/**
+ * This class is the default implementation of WebViewEventListener.
+ * It can be used with delegation pattern
+ */
+
+class DefaultWebViewEventListener : WebViewEventListener {
```
bmarty: Maybe add `@CallSuper` on each methods?

vector/src/main/java/im/vector/webview/ConsentWebViewEventListener.kt
ganfra: It's a copy fail :)

vector/src/main/java/im/vector/webview/ConsentWebViewEventListener.kt
ganfra: Nope, as we retry request until it succeeds. The user can still close the screen if it takes too much time

ganfra: Same

vector/src/main/java/im/vector/webview/DefaultWebViewEventListener.kt
ganfra: yup

vector/src/main/java/im/vector/webview/DefaultWebViewEventListener.kt
ganfra: This class is not open and shouldn't be extended, so nope :-)

PR #2502, opened 2018-08-08
Feature/fix 2409

### LINKED ISSUE #2409: In some cases, devices list is not shown in the settings
Every time I open the settings from the menu, a toast is shown, saying java.lang.Boolean cannot be cast to java.lang.String.

Screenshot of toast saying java.lang.Boolean cannot be cast to java.lang.String

bmarty:
Fix #2409: VectorCustomActionEditTextPreference are never used as an EditTextPrefence

So it now just extends Preference, so it avoid trying to cast a stored Boolean from the Prefs (why it has been stored is a mystery) to a String to display the EditText value

Main settings and Room settings (who are both using VectorCustomActionEditTextPreference) has been tested Ok

This PR also contains some cleanup

### DIFF
--- a/vector/src/main/java/im/vector/fragments/VectorRoomSettingsFragment.java
+++ b/vector/src/main/java/im/vector/fragments/VectorRoomSettingsFragment.java
@@ -353,7 +353,7 @@ public class VectorRoomSettingsFragment extends PreferenceFragment implements Sh
         });
 
         // display the room Id.
-        EditTextPreference roomInternalIdPreference = (EditTextPreference) findPreference(PREF_KEY_ROOM_INTERNAL_ID);
+        Preference roomInternalIdPreference = findPreference(PREF_KEY_ROOM_INTERNAL_ID);
         if (null != roomInternalIdPreference) {
             roomInternalIdPreference.setSummary(mRoom.getRoomId());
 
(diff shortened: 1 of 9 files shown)


vector/src/main/java/im/vector/fragments/VectorRoomSettingsFragment.java
```diff
@@ -353,7 +353,7 @@ public void onWarningIconClick(Preference preference) {
         });
 
         // display the room Id.
-        EditTextPreference roomInternalIdPreference = (EditTextPreference) findPreference(PREF_KEY_ROOM_INTERNAL_ID);
+        Preference roomInternalIdPreference = findPreference(PREF_KEY_ROOM_INTERNAL_ID);
```
ganfra: Could be final?

vector/src/main/java/im/vector/fragments/VectorRoomSettingsFragment.java
```diff
@@ -367,7 +367,7 @@ public boolean onPreferenceClick(Preference preference) {
         }
 
         // leave room
-        EditTextPreference leaveRoomPreference = (EditTextPreference) findPreference(PREF_KEY_ROOM_LEAVE);
+        Preference leaveRoomPreference = findPreference(PREF_KEY_ROOM_LEAVE);
```
ganfra: Could be final?

PR #2514, opened 2018-08-11
Don't fetch previews from server when URL previews are disabled

ldruschk:
Currently, the client will request the URL preview from the server even when displaying URL previews is disabled. This leads to the contents of E2E-encrypted chats being leaked to the homeserver without the ability to prevent this on the client.

This change sets the URLPreview to null in manageURLPreviews when URL previews are disabled in the session. Missing previews will be fetched when the switch is re-enabled later.

### DIFF
--- a/vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
+++ b/vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
@@ -1219,7 +1219,7 @@ class VectorMessagesAdapterHelper {
             final String downloadKey = url.hashCode() + "---";
             String displayKey = url + "<----->" + id;
 
-            if (!this.mSession.isURLPreviewEnabled()) {
+            if (!mSession.isURLPreviewEnabled()) {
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/adapters/VectorMessagesAdapterHelper.java
```diff
@@ -1219,7 +1219,12 @@ void manageURLPreviews(final Message message, final View convertView, final Stri
             final String downloadKey = url.hashCode() + "---";
             String displayKey = url + "<----->" + id;
 
-            if (UrlPreviewView.Companion.didUrlPreviewDismiss(displayKey)) {
+            if (!this.mSession.isURLPreviewEnabled()) {
```
bmarty: Please remove useless `this.`

ldruschk:
> When you say Missing previews will be fetched when the switch is re-enabled later. it's because the Fragment will be recreated after the settings has been updated, due to the navigation, and so mUrlsPreviews will be cleared, right?

Yeah, that's just the existing functionality and not something I implemented.

PR #2518, opened 2018-08-13
Improve managment on permission WRITE_EXTERNAL_STORAGE

bmarty:
Globally improve the management of permission on Riot for Android M + (even if it not perfect yet)


vector/src/main/java/im/vector/adapters/VectorMediasViewerAdapter.java
```diff
@@ -641,65 +643,67 @@ private void playVideo(View pageView, VideoView videoView, File videoFile, Strin
     /**
      * Download the current media file, and export it to the Download folder of the device
      */
-    private void downloadMediaAndExportToDownloads() {
-        final SlidableMediaInfo mediaInfo = mMediasMessagesList.get(mLatestPrimaryItemPosition);
+    public void downloadMediaAndExportToDownloads() {
```
ganfra: Maybe take advantage of this refactoring to put this method elsewhere ?

vector/src/main/java/im/vector/util/PermissionsTools.kt
```diff
@@ -0,0 +1,392 @@
...
+ *
+ * If a permission was already denied by the user, a popup is displayed to
+ * explain why vector needs the corresponding permission.
+ *
+ * @param permissionsToBeGrantedBitMap the permissions bit map to be granted
+ * @param activity                     the calling Activity that is requesting the permissions (or fragment parent)
+ * @param fragment                     the calling fragment that is requesting the permissions
+ * @return true if the permissions are granted (synchronous flow), false otherwise (asynchronous flow)
+ */
+private fun checkPermissions(permissionsToBeGrantedBitMap: Int,
```
ganfra: Maybe try to split the method?

vector/src/main/java/im/vector/adapters/VectorMediasViewerAdapter.java
bmarty: I thought about it, because it has nothing to do in an Adapter, but I cannot refactor the entire app on each PR ;)

PR #2580, opened 2018-09-09
move provider to main manifest, name dynamically from appid

Bubu:
This makes building riot with a different appid easier, or rather it makes it automatically installable alongside the main riot app.

Maybe the new provider authority should rather be `${applicationId}.VectorApp.provider` not 100% if that makes a difference at all.


Bubu:
Gah, why does this have conflicts now? :roll_eyes:

bmarty:
Hello @Bubu ,

Thanks for the PR, I've also noticed that, and I've already done it here: 1a4acbb54a87493621a7c5ff1264677d59b66416, (it's not yet merged).

Benoît

Bubu:
Oh. Didn't see that. Then we can close this :-).

bmarty:
Ok, thanks

PR #2616, opened 2018-09-25
Use Builder to create HomeServerConnectionConfig

bmarty:
Require https://github.com/matrix-org/matrix-android-sdk/pull/359


vector/src/main/java/im/vector/activity/LoginActivity.java
```diff
@@ -2089,11 +2092,10 @@ private HomeServerConnectionConfig getHsConfig() {
 
             try {
                 mHomeserverConnectionConfig = null;
```
ganfra: Why is there still this line? is it necessary?

bmarty: in case of exception I suppose...

PR #2619, opened 2018-09-26
Fix various issues

bmarty:
Require https://github.com/matrix-org/matrix-android-sdk/pull/362


vector/src/main/java/im/vector/view/NotificationAreaView.kt
```diff
@@ -239,7 +239,7 @@ class NotificationAreaView @JvmOverloads constructor(
      */
     private inner class ResendAllClickableSpan : ClickableSpan() {
         override fun onClick(widget: View) {
-            delegate?.deleteUnsentEvents()
+            delegate?.resendUnsentEvents()
```
ganfra: Wooooops

PR #2625, opened 2018-09-28
Fix regression on nickname completion

### LINKED ISSUE #2613: Nickname completion of displaynames is broken
Action:
* Type first three letters of someone's displayname

Expected result:
* After a few seconds, see a popup with possible completions

Actual result:
* Nothing happens

Affected versions:
* Latest lazyloading build
  * With lazyloading enabled
  * With lazyloading disabled
* Apparently the latest version released to F-Droid/Google is also affected

---

Completion of matrix IDs works as expected. (`@bla...`)

bmarty:
- Improve string format
- Fix #2613 
- Travis run unit tests

### DIFF
--- a/vector/src/test/java/im/vector/util/AutoCompletionModeTest.kt
+++ b/vector/src/test/java/im/vector/util/AutoCompletionModeTest.kt
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2018 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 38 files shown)


vector/src/test/java/im/vector/util/AutoCompletionModeTest.kt
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+package im.vector.util
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class AutoCompletionModeTest {
+    @Test
+    fun userMode_empty() {
```
ganfra: I like to have unit tests name like that : 
methodName_ParamIn_ExceptedResult

getWithText_ClassicText_ShouldReturnUserMode

PR #2636, opened 2018-10-03
Init firebase. Some users report that the application crashes

bmarty:
if this is not done is not done


bmarty:
Note that this change has been reverted, because it does not fix the issue

PR #2651, opened 2018-10-11
Improve themes management

bmarty:
There is still lots of work to do to correctly apply theme colors to the entire application, but this is a first step.

@vulet: you may be interested to check the improved status theme (mainly the status and toolbar background colors)


vulet:
This looks awesome, I like the idea of it becoming more dynamic in regard to theme styling.

PR #2656, opened 2018-10-15
Update after SDK API change

bmarty:
Request https://github.com/matrix-org/matrix-android-sdk/pull/379

### DIFF
--- a/vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
+++ b/vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
@@ -39,7 +39,6 @@ import android.text.method.LinkMovementMethod;
 import android.text.style.BackgroundColorSpan;
 import android.text.style.ClickableSpan;
 import android.text.style.ForegroundColorSpan;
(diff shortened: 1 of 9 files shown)


vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
```diff
@@ -1800,6 +1773,10 @@ private boolean isSupportedRow(MessageRow row) {
 
         boolean isSupported = VectorMessagesAdapterHelper.isDisplayableEvent(mContext, row);
 
+        if (!isSupported) {
+            Log.w(LOG_TAG, "Unsupported row. Event type: " + event.getType());
+        }
+
         if (isSupported && TextUtils.equals(event.getType(), Event.EVENT_TYPE_STATE_ROOM_MEMBER)) {
```
giomfo: Can we use here `else if` and remove check on `isSupported`?

bmarty: Yes, but I prefer to keep the algo like this, and this is just for a log

PR #2670, opened 2018-10-19
fix:status.im backgrounds

vulet:
Signed-off-by: Mike vulet@protonmail.com

### DIFF
--- a/vector/src/main/res/values/theme_status.xml
+++ b/vector/src/main/res/values/theme_status.xml
@@ -4,7 +4,6 @@
     <!-- STATUS.IM THEME COLORS -->
 
     <style name="Theme.Vector.Status" parent="Theme.AppCompat.Light">
-        <!-- Only setting the items we need to override, otherwise inheriting -->
         <!-- default background color -->
         <item name="riot_primary_background_color">@color/riot_primary_background_color_status</item>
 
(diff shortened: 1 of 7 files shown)


vulet:
Before
before
After
after

vulet:
The FAB remains indistinguishable on "Room Details" for all themes, but that is to be outside of the extent of this PR.

vulet:
[#2672](https://github.com/vector-im/riot-android/issues/2672)

vector/src/main/res/values/theme_status.xml
```diff
@@ -8,12 +8,18 @@
         <!-- default background color -->
         <item name="riot_primary_background_color">@color/riot_primary_background_color_status</item>
 
+        <!-- waiting view background -->
+        <item name="waiting_background_color">#AAAAAAAA</item>
```
bmarty: What about other elements defined in style `Theme.Vector.Light` and still missing here?
Actually the comment `<!-- Only setting the items we need to override, otherwise inheriting -->` comes from the `Black` theme, which inherit the `Dark` theme (and not directly an `AppCompat` theme). I think you can remove this comment and add the missing elements (copy them from `Theme.Vector.Light` is ok for a first step.

vector/src/main/res/values/theme_status.xml
```diff
@@ -8,12 +8,18 @@
...
+
```
bmarty: The comment `<!-- Only setting the items we need to override, otherwise inheriting -->` can be removed from this file. It comes from the `Black` theme, which inherit the `Dark` theme, and not directly an `AppCompat` theme.
Also, consider adding every missing elements form theme `Theme.Vector.Light` here, to be sure there will be no other crash in the future. Keeping the same order please :)

bmarty: The comment `<!-- Only setting the items we need to override, otherwise inheriting -->` can be removed from this file. It comes from the `Black` theme, which inherit the `Dark` theme, and not directly an `AppCompat` theme.
Also, consider adding every missing elements from theme `Theme.Vector.Light` here, to be sure there will be no other crash in the future. Keeping the same order please :)

bmarty: The comment `<!-- Only setting the items we need to override, otherwise inheriting -->` can be removed from this file. It comes from the `Black` theme, which inherit the `Dark` theme, and not directly an `AppCompat` theme.
Also, consider adding every missing elements from theme `Theme.Vector.Light` here, to be sure there will be no other crash in the future. Keeping the same order please :)

bmarty: The comment `<!-- Only setting the items we need to override, otherwise inheriting -->` can be removed from this file. It comes from the `Black` theme, which inherit the `Dark` theme, and not directly an `AppCompat` theme.
Also, consider adding every missing elements from theme `Theme.Vector.Light` here, to be sure there will be no other crash in the future. Keeping the same order please :)

vulet: Hey Benoit, thank you for the helpful tips. I believe the new commits should follow more appropriately. I saw your previous commits on making the themes more consistent, and I thought that was nice.

vulet:
Hey Benoit, thank you for the helpful tips. I believe the new commits should follow more appropriately. I saw your previous commits on making the themes more consistent, and I thought that was nice.

bmarty:
Hi @vulet ,

Sorry for the several identical messages, due to GH failure at the beginning of last week. Thanks for the update, it's ok now

Benoît

PR #2691, opened 2018-10-30
Locales management has been moved to a dedicated file

bmarty:
Kotlin code, no more AsyncTask.
`VectorApp` keeps getting exploded

### DIFF
--- a/vector/src/main/java/im/vector/settings/VectorLocale.kt
+++ b/vector/src/main/java/im/vector/settings/VectorLocale.kt
@@ -1,0 +1,251 @@
+/*
+ * Copyright 2018 New Vector Ltd
+ *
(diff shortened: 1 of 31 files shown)


vector/src/main/java/im/vector/settings/VectorLocale.kt
```diff
@@ -0,0 +1,200 @@
...
+     * Provides the current application locale
+     */
+    var applicationLocale = defaultLocale
+        private set
+
+    /**
+     * Init this object
+     */
+    fun init(context: Context) {
+        val preferences = PreferenceManager.getDefaultSharedPreferences(context)
```
ganfra: could be great to let sharedPref being provided as a dependency, to improve testability

PR #2697, opened 2018-10-31
Setting the visibility of the info area.

ma1uta:
Signed-off-by: Anatoliy Sablin <sablintolya@gmail.com>

Add the preference to setup visibility of the info area in the room.
Allowed values: "always", "messages and errors", "only errors".
Default value: "always".

"Only errors" means the info area will be displayed only when errors was occurred (for example, render tombstone, connection error).
"Messages and errors" means the info area will be hided when it is empty. If it has for example scrolling or typing icon the info area will be displayed.
"Always" means the info area will be displayed always even empty.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,7 @@ Features:
  - Add RTL support (#2376, #2271)
 
 Improvements:
- -
+ - Add a setting of the room's info area visibility.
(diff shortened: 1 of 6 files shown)


ma1uta:
Ok.
Thanks.

Issue #2436, opened 2018-07-17
Settings: avoid double negation in the user interface settings


bmarty:
Example:
Change "Hide read receipt", default to false
to "Show read receipt", default to true

Note that new String may be created to avoid bad translations
Also current user settings should be migrated properly

bmarty:
Fixed with #2723

PR #2723, opened 2018-11-11
modify some settings and update PreferencesManager

dkanada:
This is related to issue #2436 and removes double negation in some of the settings. I also grouped the keys a bit more clearly in PreferencesManager and modified a couple other strings.

I tried to slim down some of the setting titles because I think it might be better in the future to explain the settings in the description and have more concise titles. One example is the inline URL previews, the title could just be `Inline URL Previews` and the description could explain that they will only be shown if your homeserver enables the feature.

### DIFF
--- a/vector/src/main/res/values/strings.xml
+++ b/vector/src/main/res/values/strings.xml
@@ -570,13 +570,13 @@
     <string name="settings_profile_picture">Profile Picture</string>
     <string name="settings_display_name">Display Name</string>
     <string name="settings_email_address">Email</string>
-    <string name="settings_add_email_address">Add email address</string>
+    <string name="settings_add_email_address">Add Email Address</string>
     <string name="settings_phone_number">Phone</string>
-    <string name="settings_add_phone_number">Add phone number</string>
-    <string name="settings_summary_app_info_link">Application system info screen</string>
-    <string name="settings_title_app_info_link">Application info</string>
+    <string name="settings_add_phone_number">Add Phone Number</string>
+    <string name="settings_app_info_link_title">Application Info</string>
+    <string name="settings_app_info_link_summary">Show the application info in the system settings.</string>
 
-    <string name="settings_notification_privacy">Notification privacy</string>
+    <string name="settings_notification_privacy">Notification Privacy</string>
     <string name="settings_notification_privacy_warning">Riot.im runs in the background to keep your notification message content private. You can change this option to reduce battery usage.</string>
     <string name="settings_notification_privacy_normal">Normal</string>
     <string name="settings_notification_privacy_power_saving">Power-saving Options</string>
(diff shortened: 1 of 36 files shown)


dkanada:
Is VectorPreferenceCategory used often? I was trying to figure out why the Labs category always has a bold title, turns out the custom category class sets text as bold, which I think looks a tad odd.

vector/src/main/res/values/strings.xml
```diff
@@ -590,24 +590,24 @@
...
     <string name="settings_turn_screen_on">Turn the screen on for 3 seconds</string>
 
-    <string name="settings_containing_my_display_name">Msgs containing my display name</string>
-    <string name="settings_containing_my_user_name">Msgs containing my user name</string>
-    <string name="settings_messages_in_one_to_one">Msgs in one-to-one chats</string>
-    <string name="settings_messages_in_group_chat">Msgs in group chats</string>
-    <string name="settings_invited_to_room">When i\'m invited to a room</string>
-    <string name="settings_call_invitations">Call invitations</string>
-    <string name="settings_messages_sent_by_bot">Messages sent by bot</string>
+    <string name="settings_containing_my_display_name">Display Name Mentions</string>
```
bmarty: I do not like `Msgs` but your new wording are less accurate than the previous one. Consider restoring the previous strings for all this section

bmarty:
Also please fix errors reported by Travis

vector/src/main/res/values/strings.xml
```diff
@@ -570,13 +570,13 @@
     <string name="settings_profile_picture">Profile Picture</string>
     <string name="settings_display_name">Display Name</string>
     <string name="settings_email_address">Email</string>
-    <string name="settings_add_email_address">Add email address</string>
+    <string name="settings_add_email_address">Add Email Address</string>
```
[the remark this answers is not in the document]
dkanada: I just figured that using the title field implies title casing should be used as much as possible, especially if the summaries are used. This also applies to preference categories, which are used as titles, which I prefer title cased. Let me know what you decide on, I can revert these changes.

vector/src/main/res/values/strings.xml
dkanada: I agree with the use of `msgs` but I thought these changes were pretty accurate. The only one that I felt was a bit of a jump was `bot messages` but `room invitations` seems like the exact same meaning as before. If that doesn't make sense then would `call invitations` also be too vague? As for the other instances, I think the term `mention` is used in other places in the app to refer to the same thing. I could change `One-to-One Chats` to `One-to-One Chat Messages` and the same for groups but I think the use of `chat` is enough to explain the meaning.

vector/src/main/res/values/strings.xml
```diff
@@ -617,41 +617,47 @@
...
-    <string name="settings_hide_join_leave_messages">Hide join/leave messages (invites/kicks/bans unaffected)</string>
-    <string name="settings_hide_avatar_display_name_changes_messages">Hide avatar and display name changes</string>
-    <string name="settings_disable_markdown">Disable markdown formatting when sending</string>
-    <string name="settings_vibrate_on_mention">Vibrate when mentioning</string>
+    <string name="settings_inline_url_preview">Inline URL Preview</string>
+    <string name="settings_inline_url_preview_summary">Preview links within the chat when your homeserver supports this feature.</string>
+    <string name="settings_send_typing_notifs">Send Typing Notifications</string>
+    <string name="settings_send_typing_notifs_summary">Let other users know that you are typing.</string>
+    <string name="settings_send_markdown">Markdown Formatting</string>
+    <string name="settings_send_markdown_summary">Send messages with markdown when you use it for italic or bold text, among other uses.</string>
```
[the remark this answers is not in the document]
dkanada: How about `Format messages using markdown syndax before they are sent. This allows for advanced formatting such as using asterisks to display italic text.`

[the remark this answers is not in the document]
dkanada: I always try to avoid using too many special characters in short descriptions haha, its also why I am being a stickler about the title casing =D

bmarty:
> Is VectorPreferenceCategory used often? I was trying to figure out why the Labs category always has a bold title, turns out the custom category class sets text as bold, which I think looks a tad odd.

I assume VectorPreferenceCategory should be used only in `vector_room_settings_preferences.xml`. Feel free to make the labs section is classical PreferenceCategory.

Also I wanted to let you know that I've started to migrate to preference from v7 support library. It's not done yet, but it will be in a near future.

vector/src/main/res/values/strings.xml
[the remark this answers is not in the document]
bmarty: After discussion with the team, please revert these changes. We prefer only the first letter capitalised.

vector/src/main/res/values/strings.xml
bmarty: The aim of this section is to set up the notification when the received message match some rules. That's why I think `Message containing my display name` is better than `Display name mention`.

dkanada:
Signed-off-by: Dillon Kanada <dkanada@protonmail.com>

vector/src/main/res/values/strings.xml
```diff
@@ -842,8 +848,8 @@
     <string name="encryption_import_room_keys_summary">Import the keys from a local file</string>
     <string name="encryption_import_import">Import</string>
     <string name="encryption_import_enter_passphrase">Enter passphrase</string>
-    <string name="encryption_never_send_to_unverified_devices_title">Encrypt to verified devices only</string>
-    <string name="encryption_never_send_to_unverified_devices_summary">Never send encrypted messages to unverified devices from this device</string>
+    <string name="encryption_never_send_to_unverified_devices_title">Only message verified devices</string>
+    <string name="encryption_never_send_to_unverified_devices_summary">If a room has encryption enabled only send messages when all devices have been verified.</string>
```
bmarty: I think the meaning is not the same than before. Sorry I haven't seen this on my first review. Can you improve (or revert) these two wordings please?

dkanada: I thought it was more clear after the change, is my wording incorrect? This setting doesn't apply to unencrypted rooms as far as I know. Plus, the old description uses the word never which makes it a negation.

This is a hard one, `Don't never send encrypted messages` is my first thought when the setting is off with the old description. The use of `only` also makes it odd at first glance, but I left that in because I couldn't improve it.

bmarty: Yes, we think the new wording is incorrect.

dkanada: Fixed this, the only changes since yesterday are adding a period at the end of that description and matching the capitalization of `Not tagged` with the other room tag options.

bmarty: (Note that R.string.room_settings_tag_pref_no_tag is never used and could be removed)

PR #2736, opened 2018-11-16
Update verify device wording

axelsimon:
Attempt at making the text more understandable for people who haven't been playing with e2e encryption concepts for years :)

### DIFF
--- a/vector/src/main/res/values/strings.xml
+++ b/vector/src/main/res/values/strings.xml
@@ -876,8 +876,8 @@
     <string name="encryption_information_unblock">Unblacklist</string>
 
     <string name="encryption_information_verify_device">Verify device</string>
-    <string name="encryption_information_verify_device_warning">To verify that this device can be trusted, please contact its owner using some other means (e.g. in person or a phone call) and ask them whether the key they see in their User Settings for this device matches the key below:</string>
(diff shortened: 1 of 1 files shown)


vector/src/main/res/values/strings.xml
```diff
@@ -876,8 +876,8 @@
     <string name="encryption_information_unblock">Unblacklist</string>
 
     <string name="encryption_information_verify_device">Verify device</string>
-    <string name="encryption_information_verify_device_warning">To verify that this device can be trusted, please contact its owner using some other means (e.g. in person or a phone call) and ask them whether the key they see in their User Settings for this device matches the key below:</string>
-    <string name="encryption_information_verify_device_warning2">If it matches, press the verify button below. If it doesn\'t, then someone else is intercepting this device and you should probably blacklist it. In the future this verification process will be more sophisticated.</string>
+    <string name="encryption_information_verify_device_warning">A new device has connected to your account and is requesting the secret keys needed to join your private conversations from this device. To verify that this new device can be trusted, please verify whether the string of characters you see in the new device's User Settings (called the key's fingerprint) matches this string below:</string>
```
bmarty: I do not think this dialog is displayed only when another device connects to your own account.

axelsimon: Ah well this is clearly going to be a problem. We would need to separate the two cases then (and that's a bit beyond what I can do at this point).

PR #2742, opened 2018-11-25
Fix AccountCreationCaptcha error logging

slade87:
If a http error occured the logging did not actually report
the error message.


slade87:
Thanks

PR #2755, opened 2018-12-04
Keys backup

bmarty:
Little PR to handle changes introduced in https://github.com/matrix-org/matrix-android-sdk/pull/401


vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
```diff
@@ -2677,9 +2678,22 @@ class VectorSettingsPreferencesFragment : PreferenceFragment(), SharedPreference
 
                 displayLoadingView()
 
-                mSession.crypto?.importRoomKeys(data, password, object : ApiCallback<Void> {
-                    override fun onSuccess(info: Void?) {
+                mSession.crypto?.importRoomKeys(data, password, object : ApiCallback<ImportRoomKeysResult> {
+                    override fun onSuccess(info: ImportRoomKeysResult?) {
+                        if (!isAdded) {
```
ganfra: Maybe test info!=null here too

bmarty: It should not be null and it is tested after (we want the loading wheel to be hidden in a ll cases)

PR #2765, opened 2018-12-12
Fix crash due to realm introduced in MatrixSDK

bmarty:

### DIFF
--- a/vector/src/main/java/im/vector/VectorApp.java
+++ b/vector/src/main/java/im/vector/VectorApp.java
@@ -179,12 +179,6 @@ public class VectorApp extends MultiDexApplication {
     };
 
     @Override
-    protected void attachBaseContext(Context base) {
-        super.attachBaseContext(base);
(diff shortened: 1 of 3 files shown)


vector/src/main/java/im/vector/VectorApp.java
```diff
@@ -193,6 +188,8 @@ public void onCreate() {
             Stetho.initializeWithDefaults(this);
         }
 
+        Realm.init(this);
```
ganfra: Why do we need that here ? It should be hidden, isn't it?

bmarty: yes, but why not? the init() will do think only once. And maybe we will use Realm in Riot for other usage... Do you prefer me to remove this call here?

ganfra: I don't want to see Realm in Riot if it's not used directly yeah...

PR #2790, opened 2018-12-20
Fix UI/UX issues

### LINKED ISSUE #2183: Show what info rageshake is going to send
When I rageshake my phone and file a bug report, I'm asked to tick what I'm going to send. A screenshot, logs, crash logs. I don't know what's in the logs, I want to read them before sending! And the screenshot. Wanna make sure there are no bugs making my private rooms' screenshots sent instead of what I wanted.

The system is nice, but should be a little more transparent.
Thanks!

Sorry if this is a duplicate, couldn't find anything related.

### LINKED ISSUE #2769: UI/UX issue for Sprint 17
- [x] **Detail member screen -> device list**: UI is not refreshed when blacklist/unblacklist a device
- [x] **Room detail screen**: link in the topic are not visible (same color of the background). Also increase the size of this TextView (too much margin)
- [x] **Room detail screen**: "Add a new address" item is truncated. Should be displayed on two lines
- [x] **Register screen**: "Use custom server option" is not align correctly 
- [x] **Room screen**: When downloading media, the cross is always black, whatever the theme is
- [x] cross to close url preview is always green. Bad for status theme
- [x] **Send attachment popup** bad rendering with RTL language
- [x] **Global Search/Historical search** hint is not white in SearchView
- [x] **Global search** RTL issue on arrow on left of group screen

bmarty:
Fix #2769 
Fix #2183

### DIFF
--- a/vector/src/main/res/layout/adapter_item_member_details_devices.xml
+++ b/vector/src/main/res/layout/adapter_item_member_details_devices.xml
@@ -33,8 +33,7 @@
         android:layout_alignParentRight="true"
         android:layout_marginEnd="16dp"
         android:layout_marginRight="16dp"
-        android:src="@drawable/e2e_verified"
-        android:tint="?attr/vctr_default_icon_tint_color" />
+        tools:src="@drawable/e2e_verified" />
 
     <LinearLayout
         android:id="@+id/container_buttons"
--- a/vector/src/main/res/layout/adapter_item_member_details_devices.xml
+++ b/vector/src/main/res/layout/adapter_item_member_details_devices.xml
@@ -8,21 +8,21 @@
 
     <TextView
         android:id="@+id/device_name"
-        android:layout_width="match_parent"
+        android:layout_width="wrap_content"
         android:layout_height="wrap_content"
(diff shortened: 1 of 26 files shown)


vector/src/main/res/layout/adapter_item_member_details_devices.xml
```diff
@@ -8,21 +8,21 @@
...
         android:layout_height="wrap_content"
         android:layout_below="@+id/device_name"
         android:layout_marginStart="16dp"
         android:layout_marginLeft="16dp"
         android:layout_marginTop="0dp"
-        tools:text="device id.." />
+        tools:text="device id" />
 
     <ImageView
         android:id="@+id/device_e2e_icon"
```
BillCarsonFr: I noticed that a tint color is applied to this image. Thus the warning or 'no entry' icon are green. The color for these icon carries semantic and should not be tinted.

bmarty: ok, done. I've also implemented the ViewHolder pattern in the Adapter

PR #2803, opened 2018-12-28
Stopping LoadingView after UserAvatar Upload

theLimitBreaker:
### Pull Request Checklist

* :heavy_check_mark: Pull request is based on the develop branch
* :heavy_check_mark: Pull request updates [CHANGES.rst](https://github.com/vector-im/riot-android/blob/develop/CHANGES.rst)
* :heavy_check_mark: Pull request includes screenshots or videos if containing UI changes
* :heavy_check_mark: Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

Preview


vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
```diff
@@ -1377,7 +1379,7 @@ class VectorSettingsPreferencesFragment : PreferenceFragmentCompat(), SharedPref
                                         mSession.myUser.updateAvatarUrl(contentUri, object : ApiCallback<Void> {
                                             override fun onSuccess(info: Void?) {
                                                 onCommonDone(null)
-                                                refreshDisplay()
+                                                hideLoadingView(true)
```
bmarty: This is already done in `onCommonDone()` so I think this can be removed.

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
```diff
@@ -870,8 +870,10 @@ class VectorSettingsPreferencesFragment : PreferenceFragmentCompat(), SharedPref
     override fun onResume() {
         super.onResume()
 
-        // search the loading view from the upper view
-        mLoadingView = view!!.findViewById(R.id.vector_settings_spinner_views)
+        // search the loading view from the upper
```
bmarty: This comment can be remove.

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
```diff
@@ -870,8 +870,10 @@ class VectorSettingsPreferencesFragment : PreferenceFragmentCompat(), SharedPref
...
+        // find the view from parent activity
+        mLoadingView = activity!!.findViewById(R.id.vector_settings_spinner_views)
```
bmarty: Note that I really don't like the fact that a Fragment manipulates View from its Activity, but the code is full of dirty practices like that, so it is out of scope of this PR...

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
theLimitBreaker: onCommonDone() calls just hideLoadingView() doesn't refreshDisplay(). So I see , we can call refreshDisplay() only

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
theLimitBreaker: So should I move the loading view to fragment only. ?

bmarty: No, You won't be able to do that because it's a pref fragment. I was more thinking of a way for Fragment to ask its Activity to show/hide the loading View.

theLimitBreaker: We can have Interface Call back. VectorSettingPreferencesFragment Can have interface for displayLoadingView and HideLoadingView(). actvity can give the implementation at the time of initialization of this fragment.

theLimitBreaker: So In Kotlin Code Cleanup . Should I make this Change too ?

Issue #1262, opened 2017-06-01
Riot got into a startup loop after an OOM crash.


ylecollen:
Reported by Coffee

Riot got into a startup loop after an OOM crash. Had to force-stop it to come out of it. 

https://github.com/matrix-org/riot-android-rageshakes/issues/217

PR #2820, opened 2019-01-03
Restore the crash report management

bmarty:
It has been removed by mistake when Google Analytics has been removed from Riot

@giomfo you may be interested to disable this on your favorite project :)

### DIFF
--- a/vector/src/main/java/im/vector/activity/VectorHomeActivity.java
+++ b/vector/src/main/java/im/vector/activity/VectorHomeActivity.java
@@ -126,6 +126,7 @@ import im.vector.fragments.RoomsFragment;
 import im.vector.push.PushManager;
 import im.vector.receiver.VectorUniversalLinkReceiver;
 import im.vector.services.EventStreamService;
+import im.vector.tools.VectorUncaughtExceptionHandler;
 import im.vector.ui.themes.ActivityOtherThemes;
 import im.vector.ui.themes.ThemeUtils;
 import im.vector.util.BugReporter;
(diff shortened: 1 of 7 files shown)


vector/src/main/java/im/vector/services/EventStreamService.java
```diff
@@ -353,7 +354,7 @@ public int onStartCommand(Intent intent, int flags, int startId) {
                     return START_NOT_STICKY;
                 }
 
-                if ((null != VectorApp.getInstance()) && VectorApp.getInstance().didAppCrash()) {
+                if (VectorUncaughtExceptionHandler.INSTANCE.didAppCrash(this)) {
```
BillCarsonFr: Does that mean that notifications won't work until the app is manually restarted?

vector/src/main/java/im/vector/activity/VectorHomeActivity.java
```diff
@@ -573,7 +575,7 @@ public void onClick(DialogInterface dialog, int which) {
                         })
                         .show();
 
-                VectorApp.getInstance().clearAppCrashStatus();
+                VectorUncaughtExceptionHandler.INSTANCE.clearAppCrashStatus(this);
```
BillCarsonFr: i think this should be cleared just after the didCrash test or in a finally?

bmarty: Yes, TBH, I do not understand this try catch block, and I'm really tempted to remove it

vector/src/main/java/im/vector/services/EventStreamService.java
bmarty: I assume yes, according to git history, it was added to fix https://github.com/vector-im/riot-android/issues/1262

vector/src/main/java/im/vector/activity/VectorHomeActivity.java
bmarty: Should be interesting to search in (old) crash logs if `appCrashedAlert failed` occurred

vector/src/main/java/im/vector/services/EventStreamService.java
BillCarsonFr: Maybe we should display a notification in this case? 'Oups, sorry but Riot did crash. In order to maintain your system integrity Riot will not restart background activities until you re-start the app manually, this might affect notification reliability'

bmarty: Yes, maybe, but I think we can do this in a dedicated PR

PR #2821, opened 2019-01-04
Disable usage of library ShortcutBadger on device running API 26+

bmarty:

### DIFF
--- a/vector/src/main/java/im/vector/ui/badge/BadgeProxy.kt
+++ b/vector/src/main/java/im/vector/ui/badge/BadgeProxy.kt
@@ -1,0 +1,133 @@
+/*
+ * Copyright 2019 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


vector/src/main/java/im/vector/ui/badge/BadgeProxy.kt
```diff
@@ -0,0 +1,135 @@
...
+ */
+object BadgeProxy {
+    private const val LOG_TAG = "BadgeProxy"
+
+    /**
+     * Badge is now managed by notification channel, so no need to use compatibility library in recent versions
+     *
+     * @return true if library ShortcutBadger can be used
+     */
+    private fun useShortcutBadger(): Boolean {
```
BillCarsonFr: use short syntax

vector/src/main/java/im/vector/ui/badge/BadgeProxy.kt
```diff
@@ -0,0 +1,135 @@
...
+        return Build.VERSION.SDK_INT < Build.VERSION_CODES.O
+    }
+
+    /**
+     * Update the application badge value.
+     *
+     * @param context    the context
+     * @param badgeValue the new badge value
+     */
+    fun updateBadgeCount(context: Context, badgeValue: Int) {
```
BillCarsonFr: I think we might need to update notification channels default settings to state if we want to display or not badges (setShowBadge(true/false)).
And will probably need some work on notification builder with .setNumber(messageCount)

bmarty: Yes

vector/src/main/java/im/vector/ui/badge/BadgeProxy.kt
bmarty: Done

Issue #2854, opened 2019-01-10
Anonymous AsyncTask . Bad CodeStyle . Potential Memory Leak


theLimitBreaker:
In VectorPreferenceFragment , for clearing media cache, there is potential memory leak.

`val task = object : AsyncTask<Void?, Void?, Void?>() {
                    override fun doInBackground(vararg params: Void?): Void? {
                        mSession.mediaCache.clear()
                        Glide.get(activity!!).clearDiskCache()
                        return null
                    }

                    override fun onPostExecute(result: Void?) {
                        hideLoadingView()

                        MXMediaCache.getCachesSize(activity, object : SimpleApiCallback<Long>() {
                            override fun onSuccess(size: Long) {
                                it.summary = android.text.format.Formatter.formatFileSize(activity, size)
                            }
                        })
                    }
                }`

**Solution** 
There should be separate ClearMediaAsyncTask with WeakReferences of objects. So that there wont be no potential memory leak.

PR #2857, opened 2019-01-11
Kotlin Code Cleanup in VectorSettingsPreferencesFragment

theLimitBreaker:
### Pull Request Checklist

* [ :heavy_check_mark: ] Changes has been tested on an Android device or Android emulator with API 16
* [ :heavy_check_mark: ] Pull request is based on the develop branch
* [ :heavy_check_mark: ] Pull request updates [CHANGES.rst](https://github.com/vector-im/riot-android/blob/develop/CHANGES.rst)
* [ :heavy_check_mark:  ] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

Kotlin Code improvement in VectorSettingsPreferenceFragment
Remove unnecessary !! 
Replace !! with null safe operators ( ?.let{..} )
Add Separate ClearMediaAsyncTask to avoid potential memory leak (#2854)


theLimitBreaker:
@bmarty  Can you please review this ?

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
```diff
@@ -409,13 +412,51 @@ class VectorSettingsPreferencesFragment : PreferenceFragmentCompat(), SharedPref
                         onPushRuleClick(preference.key, newValueAsVoid as Boolean)
                         true
                     }
+
+                } else if (preference is BingRulePreference) {
```
bmarty: This code has been remove on /develop, please remove it

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
```diff
@@ -725,26 +778,24 @@ class VectorSettingsPreferencesFragment : PreferenceFragmentCompat(), SharedPref
...
-                    override fun doInBackground(vararg params: Void?): Void? {
-                        mSession.mediaCache.clear()
-                        Glide.get(activity!!).clearDiskCache()
-                        return null
-                    }
-
-                    override fun onPostExecute(result: Void?) {
-                        hideLoadingView()
+                val task = ClearMediaCacheAsyncTask(
+                        {
```
bmarty: maybe add named parameters for clarity

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
```diff
@@ -921,20 +931,16 @@ class VectorSettingsPreferencesFragment : PreferenceFragmentCompat(), SharedPref
                 mLoadingView = parent.findViewById(R.id.vector_settings_spinner_views)
                 parent = parent.parent as View
             }
-        }
-
-        if (null != mLoadingView) {
-            mLoadingView!!.visibility = View.VISIBLE
+        } else {
+            mLoadingView?.visibility = View.VISIBLE
```
bmarty: As a remark, the code is not equivalent as before, because in the case where `mLoadingView` it will not be set to visible.
That said, it should never happen because `mLoadingView` is retrieved in `onResume()`

bmarty:
Thanks or the update :)

vector/src/main/java/im/vector/fragments/VectorSettingsPreferencesFragment.kt
theLimitBreaker: I understand. But there was case that , if it's null , then it would find `mLoadingView` from parent.

But Kotlin compiler wasn't smart enough for that . I was thinking using Kotlin Contracts

bmarty: Yes, but when found, the visibility will not be set to visible (because it's the else block now).
Anyway, it's ok to leave it like that

theLimitBreaker: When I do the work for Coroutines in CleanAsyncTask. I can do the work for Contracts . If want. Or else would just do for Coroutine .

PR #2858, opened 2019-01-11
Fix registration issue

bmarty:
This PR also improve the layout and the style management of LoginActivity, especially regarding i18n

It seems that sending `bind_msisdn` and `bind_email` during the terms acceptation step with no `username` and `password` makes the server forget at least the password

### DIFF
--- a/vector/src/main/java/im/vector/activity/LoginActivity.java
+++ b/vector/src/main/java/im/vector/activity/LoginActivity.java
@@ -2359,7 +2359,6 @@ public class LoginActivity extends MXCActionBarActivity implements RegistrationM
         mRegisterPollingRunnable = new Runnable() {
             @Override
             public void run() {
-                // TODO Cancel the handler when activity is paused (and not destroy)
                 Log.d(LOG_TAG, "## onWaitingEmailValidation attempt registration");
                 mRegistrationManager.attemptRegistration(LoginActivity.this, LoginActivity.this);
                 mHandler.postDelayed(mRegisterPollingRunnable, REGISTER_POLLING_PERIOD);
--- a/vector/src/main/java/im/vector/activity/LoginActivity.java
+++ b/vector/src/main/java/im/vector/activity/LoginActivity.java
@@ -115,12 +115,9 @@ public class LoginActivity extends MXCActionBarActivity implements RegistrationM
     private static final int MODE_ACCOUNT_CREATION_THREE_PID = 5;
 
     // saved parameters index
-
     // creation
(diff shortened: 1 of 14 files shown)


vector/src/main/java/im/vector/activity/LoginActivity.java
```diff
@@ -1858,22 +1817,6 @@ public void onMatrixError(MatrixError e) {
     private void restoreSavedData(@NonNull Bundle savedInstanceState) {
         Log.d(LOG_TAG, "## restoreSavedData(): IN");
 
-        mLoginEmailTextView.setText(savedInstanceState.getString(SAVED_LOGIN_EMAIL_ADDRESS));
-        mLoginPasswordTextView.setText(savedInstanceState.getString(SAVED_LOGIN_PASSWORD_ADDRESS));
-        mUseCustomHomeServersCheckbox.setChecked(savedInstanceState.getBoolean(SAVED_IS_SERVER_URL_EXPANDED));
-        mHomeServerText.setText(savedInstanceState.getString(SAVED_HOME_SERVER_URL));
```
giomfo: Are you sure you want to lose the server config when the app is backgrounded on a device where the background process limit is 0?

vector/src/main/java/im/vector/activity/LoginActivity.java
```diff
@@ -1858,22 +1817,6 @@ public void onMatrixError(MatrixError e) {
...
-        mIdentityServerText.setText(savedInstanceState.getString(SAVED_IDENTITY_SERVER_URL));
-
-        mCreationUsernameTextView.setText(savedInstanceState.getString(SAVED_CREATION_USER_NAME));
-        mCreationPassword1TextView.setText(savedInstanceState.getString(SAVED_CREATION_PASSWORD1));
-        mCreationPassword2TextView.setText(savedInstanceState.getString(SAVED_CREATION_PASSWORD2));
-
-        mForgotEmailTextView.setText(savedInstanceState.getString(SAVED_FORGOT_EMAIL_ADDRESS));
-        mForgotPassword1TextView.setText(savedInstanceState.getString(SAVED_FORGOT_PASSWORD1));
-        mForgotPassword2TextView.setText(savedInstanceState.getString(SAVED_FORGOT_PASSWORD2));
```
giomfo: Same question about the form fields

giomfo: ok I was wrong,  EditTexts contents are saved automatically by the System

Issue #2867, opened 2019-01-15
Room Version Upgrade: Clicking the link on the tombstone event invites you to choose which app you want to load the link in


lampholder:
![image](https://user-images.githubusercontent.com/1922197/51198021-01e9ee00-18eb-11e9-8c15-f42e89b08732.png)

![image](https://user-images.githubusercontent.com/1922197/51198028-06160b80-18eb-11e9-9b09-c8b4599f402e.png)

bmarty:
It's working for me, I'm redirected to the next room preview where I can join the room.

ganfra:
This link is handled as a permalink. Do you have the same process when clicking on a permalink in a message?

lampholder:
No, and if I _do_ allow the link to open in Chrome, copy the URL (of the form https://matrix.to/#/!RoOmId:matrix.org), paste it the composer and send it into the timeline, then click on the link in the timeline, clicking _that_ link works fine.

So it's only clicking the link in the tombstone event that invites me to open it in the application of my choice.

PR #2873, opened 2019-01-16
Quick fix for tombstone events to work as expected...

ganfra:
This is fixing issues #2867 and #2867


vector/src/main/java/im/vector/activity/VectorRoomActivity.java
```diff
@@ -668,6 +668,40 @@ public boolean onTouch(View view, MotionEvent motionEvent) {
...
+            @Override
+            public void jumpToBottom() {
+                if (mReadMarkerManager != null) {
+                    mReadMarkerManager.handleJumpToBottom();
+                } else {
+                    mVectorMessageListFragment.scrollToBottom(0);
+                }
+            }
+        });
+        Log.d(LOG_TAG, "End of create");
```
bmarty: Please move this log back to the end of file

PR #2882, opened 2019-01-20
remove long click download action

### LINKED ISSUE #2868: minor improvements to media viewer
I would like to remove the download on long click action because it conflicts with a really nice android gesture (double tap and drag up and down to zoom) and I would also like to add an action to download media to a specific folder.

@bmarty any opinions on the two changes before I start working on them?

dkanada:
This fixes #2868 and reduces duplicate code.


vector/src/main/java/im/vector/activity/VectorMediaViewerActivity.java
```diff
@@ -338,9 +338,6 @@ public void onRequestPermissionsResult(int requestCode, @NonNull String[] permis
             if (requestCode == PermissionsToolsKt.PERMISSION_REQUEST_CODE) {
                 // Request comes from here
                 onAction(mPendingPosition, mPendingAction);
-            } else if (requestCode == PermissionsToolsKt.PERMISSION_REQUEST_OTHER) {
```
bmarty: `PERMISSION_REQUEST_OTHER` could be removed too.

dkanada: Do you mean `PERMISSION_REQUEST_CODE` in the line above? I think that might be used for the normal download action.

bmarty: No, I mean removing `PERMISSION_REQUEST_OTHER` from file `PermissionsTools.kt`, sorry, I was not clear

dkanada:
@bmarty should be good.

bmarty:
Yes, thanks

PR #2923, opened 2019-02-05
Fix warning "Attribute value must be constant"

Legi429:
- Id starts with "R.id." as described in @BindView docs

Signed-off-by: Marco Festini <marco.festini@awesome-technologies.de>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,7 @@ Other changes:
  -
 
 Bugfix:
- -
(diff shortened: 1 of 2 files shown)


bmarty:
Also where have you seen this warning?

Legi429:
> LGTM, I assume you have checked that there is no regression?

I haven't seen any effect other than it removing the warning.

> Also where have you seen this warning?

Right in Android Studio. Showed up as a red line in the scroll bar.
The project was able to build with or without the fix.

Issue #2898, opened 2019-01-25
Settings: Password ******** is missleading


alexte:
The settings dialog looks as if the riot client remembers the password "passsword: *******"
And clicking this would change the password in the client used to login.

But actually, it changes the password on the server.

I would change this text to "Password: Change password" or "Change password on server" 

The reason I found this is, that users asked me  where they could change their password.

theLimitBreaker:
Also When you by mistake enter wrong old password or you mismatch passwords , its just says "Failed to update password". I am doing this

PR #2936, opened 2019-02-09
Make Change password settings more user friendly

theLimitBreaker:
Referencing #2898 
better error message when user enters wrong password
More User friendly UI when checking "New Password" and "Confirm New Password"

Signed-off-by: thelimitbreaker <rohanmaityofficial@gmail.com>

### Pull Request Checklist

* [:heavy_check_mark:] Changes has been tested on an Android device or Android emulator with API 16
* [:heavy_check_mark: ] Pull request is based on the develop branch
* [:heavy_check_mark: ] Pull request updates [CHANGES.rst](https://github.com/vector-im/riot-android/blob/develop/CHANGES.rst)
* [ :heavy_check_mark:] Pull request includes screenshots or videos if containing UI changes
* [ :heavy_check_mark:] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)


bmarty:
Hi @theLimitBreaker ,

Thanks for the PR.

Please set the "hard wrap" setting of Android Studio to 160 chars, to avoid extra new line when formatting source code (I will add this to the contributing guide).

A better option to handle mismatch of new passwords should be to use TextInputLayout and to display the error inline in the dialog only when the user select "save" (which should be updated to a better wording "update password").

Also I do not like the wording "Change password on server". "Change password" is better. But in this case, why not changing "Display Name" to "Change display Name"? That's a reason we should keep the legacy wording "Display Name" and "Password". I think it's quite obvious that clicking on the item will aim to update it.

And you also have to check the behavior on dark theme (I will add this on PR checks).

Benoît

theLimitBreaker:
Ok. So these are the changes I have to make
1) making use of TextInputlayout 
2) making use of Legacy wording "Display Name" and "Password" . 

and then re submit PR

theLimitBreaker:
@bmarty please check these. Once done . I would move to issue "No visual difference in text disabled and enabled"

# Working demo (theme Status.im)



# Dark theme


# Light Theme

theLimitBreaker:
>display the error inline in the dialog only when the user select "save"

Users want instantaneous response
Displaying that passwords do not match before hitting save/update button is more interactive and immediate response to user.

theLimitBreaker:
@bmarty  I accidently deleted the develop remotr branch. I have made new PR #2986 . Checkout that one

PR #2962, opened 2019-02-18
Show direct chat section in user details only for other users, not self

Legi429:
Since directs chats with myself don't make much sense, I thought it's a reasonable change to not show this section in the user details activity.

screenshot_1550499080


bmarty:
Hi @Legi429 ,

Yes, indeed, good idea!

Benoît

ptman:
I use chat with myself as a way to move stuff easily between mobile and desktop (like pushbullet), but a single-person unlisted room works as well as direct chat.

Issue #2999, opened 2019-02-25
KeysBackup issues for Sprint 19


bmarty:
- [ ] Wording "our server" -> "your Homeserver"
- [ ] link to the FAQ (where?)
- [ ] integrate StepProgress API break from SDK when importing keys

PR #3000, opened 2019-02-25
KeysBackup: Improve UX

bmarty:
Require https://github.com/matrix-org/matrix-android-sdk/pull/437

Improve UX when restoring keysbackup.

Introduce WaitingViewData class to manage the WaitingView model

Fixes partially #2999

### DIFF
--- a/vector/src/main/java/im/vector/activity/SimpleFragmentActivity.kt
+++ b/vector/src/main/java/im/vector/activity/SimpleFragmentActivity.kt
@@ -15,14 +15,14 @@
  */
 package im.vector.activity
 
-import android.support.transition.TransitionManager
-import android.view.ViewGroup
+import android.view.View
 import android.widget.ProgressBar
 import android.widget.TextView
 import androidx.core.view.isGone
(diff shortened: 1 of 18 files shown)


vector/src/main/java/im/vector/activity/SimpleFragmentActivity.kt
```diff
@@ -51,18 +50,29 @@ open class SimpleFragmentActivity : MXCActionBarActivity() {
      * Displays a progress indicator with a message to the user.
      * Blocks user interactions.
      */
-    fun showWaitingView(status: String) {
-        waitingStatusText.text = status
-        waitingStatusProgress.isVisible = false
-        showWaitingView()
-    }
+    fun updateWaitingView(data: WaitingViewData?) {
```
BillCarsonFr: no more animation on status update?

bmarty: No, but the dialog width is now set to 75% of the screen in portrait and 50% in Landscape, to limit dialog size change

Issue #2982, opened 2019-02-21
.well-known support


manuroe:
@manuroe commented on [Wed Dec 19 2018](https://github.com/vector-im/riot-meta/issues/243)

https://matrix.org/docs/spec/client_server/r0.4.0.html#server-discovery

JS PR:  matrix-org/matrix-react-sdk#2227

---

@turt2live commented on [Wed Dec 19 2018](https://github.com/vector-im/riot-meta/issues/243#issuecomment-448574815)

There's more documentation on the implementation here: https://github.com/matrix-org/matrix-js-sdk/pull/799

PR #3001, opened 2019-02-26
 .well-known support #2982

BillCarsonFr:
Require https://github.com/matrix-org/matrix-android-sdk/pull/436

Autodiscovery is performed when the userLogin field loses focus and when the field value is a MX id (@xxx:<domain>)

### DIFF
--- a/vector/src/main/java/im/vector/activity/LoginActivity.java
+++ b/vector/src/main/java/im/vector/activity/LoginActivity.java
@@ -51,6 +51,7 @@ import android.widget.Toast;
 
 import org.matrix.androidsdk.HomeServerConnectionConfig;
 import org.matrix.androidsdk.MXSession;
+import org.matrix.androidsdk.login.AutoDiscovery;
 import org.matrix.androidsdk.rest.callback.ApiCallback;
 import org.matrix.androidsdk.rest.callback.SimpleApiCallback;
 import org.matrix.androidsdk.rest.client.LoginRestClient;
@@ -338,6 +339,8 @@ public class LoginActivity extends MXCActionBarActivity implements RegistrationM
 
     // use to reset the password when the user click on the email validation
     private ThreePidCredentials mForgotPid = null;
+
+    Boolean shouldAutodiscoverDomainBasedOnMail = false;
 
     // network state notification
     private final BroadcastReceiver mNetworkReceiver = new BroadcastReceiver() {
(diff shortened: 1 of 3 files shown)


vector/src/main/java/im/vector/activity/LoginActivity.java
```diff
@@ -672,6 +675,86 @@ public void afterTextChanged(Editable s) {
 
         mTextInputLayouts = ViewUtilKt.findAllTextInputLayout(mFormContainer);
         ViewUtilKt.autoResetTextInputLayoutErrors(mTextInputLayouts);
+
+
+        mLoginEmailTextView.setOnFocusChangeListener((v, hasFocus) -> {
+            if (hasFocus) return;
```
bmarty: I'm wondering if this is called when we switch to create account mode (?)

vector/src/main/java/im/vector/activity/LoginActivity.java
```diff
@@ -339,6 +340,8 @@
     // use to reset the password when the user click on the email validation
     private ThreePidCredentials mForgotPid = null;
 
+    Boolean shouldAutodiscoverDomainBasedOnMail = false;
```
bmarty: I would remove this code and all related to this.

vector/src/main/java/im/vector/activity/LoginActivity.java
```diff
@@ -672,6 +675,86 @@ public void afterTextChanged(Editable s) {
...
         mTextInputLayouts = ViewUtilKt.findAllTextInputLayout(mFormContainer);
         ViewUtilKt.autoResetTextInputLayoutErrors(mTextInputLayouts);
+
+
+        mLoginEmailTextView.setOnFocusChangeListener((v, hasFocus) -> {
+            if (hasFocus) return;
+            //Maybe debounce
+            String candidate = mLoginEmailTextView.getText().toString();
+            //Try to see if we can find a domain
+            if (!candidate.isEmpty() && candidate.startsWith("@") && candidate.contains(":")) {
```
bmarty: Use `MXPatterns.isUserId()`

vector/src/main/java/im/vector/activity/LoginActivity.java
BillCarsonFr: Tested, it's not called (it's another text view for register)

vector/src/main/java/im/vector/activity/LoginActivity.java
BillCarsonFr: ok

vector/src/main/java/im/vector/activity/LoginActivity.java
BillCarsonFr: done

vector/src/main/java/im/vector/activity/LoginActivity.java
```diff
@@ -672,6 +674,87 @@ public void afterTextChanged(Editable s) {
...
         mTextInputLayouts = ViewUtilKt.findAllTextInputLayout(mFormContainer);
         ViewUtilKt.autoResetTextInputLayoutErrors(mTextInputLayouts);
+
+
+        mLoginEmailTextView.setOnFocusChangeListener((v, hasFocus) -> {
+            if (hasFocus) return;
+            //Maybe debounce
+            String candidate = mLoginEmailTextView.getText().toString();
+            //Try to see if we can find a domain
+            if (!candidate.isEmpty() && MXPatterns.isUserId(candidate)) {
```
bmarty: canditate.isEmpty is not necessary, isUserId will return false

BillCarsonFr:
:/
autcomplete

vector/src/main/java/im/vector/activity/LoginActivity.java
bmarty: ok

PR #3014, opened 2019-03-01
Fix typo in catalan translation

tsdgeos:

### DIFF
--- a/vector/src/main/res/values-ca/strings.xml
+++ b/vector/src/main/res/values-ca/strings.xml
@@ -458,7 +458,7 @@ Tingueu en compte que aquesta acció reiniciarà l\'aplicació i que pot trigar 
     <string name="settings_app_info_link_summary">Mostra la informació de l\'aplicació als paràmetres del sistema.</string>
     <string name="settings_app_info_link_title">Informació de l\'aplicació</string>
 
-    <string name="settings_notification_ringtone">So de les notificación</string>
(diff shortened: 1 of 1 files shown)


tsdgeos:
@jmallach is this going to work or do i have to use that weird webapp for translating?

jmallach:
I have no idea, may be safest to do it in weblate. I can do it if you like.

tsdgeos:
@jmallach If you think that's best, lets go for it

tsdgeos:
@jmallach if you haven't fixed it yet, found another typo

 <string name="directory_search_results_title">Navega pel directpri</string>

directpri → directori

bmarty:
Hi @tsdgeos ,
Thanks for the fix. I can accept PR to fix small typos in translation, but the normal way is to use Weblate. I assume it will trigger notification to other translators about the modification they can then check.
You can commit again to fix @tsdgeos suggestions if you think i's ok.
I'll then merge this PR
Thanks
Benoit

tsdgeos:
@bmarty not really interested in having yet another login in yet another system and learning to use yet another tool when i could just fix it in the source code.

If that's not valid, as i said i'll be really happy if @jmallach does the fix via weblate.

jmallach:
I have corrected these two typos in weblate. Many thanks, @tsdgeos!

bmarty:
Fix will come with next Weblate merge

PR #3043, opened 2019-03-11
Fix / m.notice autolink not working

BillCarsonFr:
And also perform autolink on formatted body, but keep formatted span in case of overlap with an autolink one

### DIFF
--- a/vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
+++ b/vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
@@ -1238,7 +1238,7 @@ public class VectorMessagesAdapter extends AbstractMessagesAdapter {
                 bodyTextView.setText(result);
 
                 mHelper.applyLinkMovementMethod(bodyTextView);
-                VectorLinkifyKt.vectorCustomLinkify(bodyTextView);
+                VectorLinkifyKt.vectorCustomLinkify(bodyTextView, true);
                 textViews = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


bmarty:
Fixes #3043

vector/src/main/java/im/vector/adapters/VectorMessagesAdapter.java
```diff
@@ -1447,6 +1447,10 @@ private View getNoticeRoomMemberView(final int viewType, final int position, Vie
                 MatrixURLSpan.refreshMatrixSpans(strBuilder, mVectorMessagesAdapterEventsListener);
                 mHelper.applyLinkMovementMethod(noticeTextView);
                 noticeTextView.setText(strBuilder);
+                //In room member we don't want autolink, but do it for m.notice
+                if (viewType == ROW_TYPE_NOTICE) {
```
bmarty: I think the Javadoc of this method could be improved...

vector/src/main/java/im/vector/util/VectorLinkify.kt
```diff
@@ -27,15 +27,24 @@ import java.util.*
 /**
  * Better support for auto link than the default implementation
  */
-fun TextView.vectorCustomLinkify() {
+fun TextView.vectorCustomLinkify(keepExistingUrlSpan: Boolean = false) {
```
bmarty: Consider adding unit test with parameter set to `true`?

bmarty:
Thanks for the update

PR #3155, opened 2019-05-06
FDroid permanent background service

### LINKED ISSUE #3120: notifications.md is outdated
After the recent riot-android changes how background sync and notifications are handled the https://github.com/vector-im/riot-android/blob/develop/docs/notifications.md#fallback-mode-2 section is outdated.

CC @BillCarsonFr

### LINKED ISSUE #3122: Notification [Fdroid] Add a new mode for notification
New android versions have greatly limited usage of resources while in background, in order to save battery/memory usage.

In order to comply with this new model and good app behavior, the F-Droid app version is using WorkManager to schedule background sync. This mean that even if there is a guaranty that sync with home server will be performed, depending of the phone status and resource the sync can be deferred.

Some users still require that the sync are performed at a precise time in order to guaranty more real-time notifications.

A new option could be proposed to user of the Fdroid version, they could choose between:
- Save as much battery as possible
- Have notification ASAP (at the cost of battery/radio) [sync delay configurable]

This mode could require battery optim whitelisting to always keep a background service alive, or a permanent foreground notification. TBD

BillCarsonFr:
Fixes #3122

* [ ] Update Notification.md doc with up to date informations
* [ ] Finalize text copy for new options
* [x] Do not use foreground service as it is buggy and affected by doze mode in several os/devices versions/make/model -> imply to rely on battery optimization whitelisting
* [ ] Settings are too complex, try to provide smart default? (and maybe a way for advanced user to configure things?
* [ ] Re-organize notification settings screen
* [ ] Changes has been tested on an Android device or Android emulator with API 16
* [ ] Add a self diagnostic screen that allows to check the periodicity of syncs? (to help notice if/when the app stopped to sync)
* [x] UI change has been tested on both light and dark themes
* [x] Pull request is based on the develop branch
* [x] Pull request updates [CHANGES.rst](https://github.com/vector-im/riot-android/blob/develop/CHANGES.rst)
* [x] Pull request includes screenshots or videos if containing UI changes
screen1
screen2
Screen3


vector/src/main/java/im/vector/VectorApp.java
```diff
@@ -405,8 +405,9 @@ private void suspendApp() {
         for (MXSession session : sessions) {
             if (session.isAlive()) {
                 session.setIsOnline(false);
-                session.setSyncDelay(pushManager.isBackgroundSyncAllowed() ? pushManager.getBackgroundSyncDelay() : 0);
-                session.setSyncTimeout(pushManager.getBackgroundSyncTimeOut());
+                //This is manage by the event stream service
```
giomfo: These commented lines should be removed

vector/src/main/java/im/vector/VectorApp.java
```diff
@@ -523,8 +524,9 @@ private void stopActivityTransitionTimer() {
             for (MXSession session : sessions) {
                 session.getMyUser().refreshUserInfos(null);
                 session.setIsOnline(true);
-                session.setSyncDelay(0);
-                session.setSyncTimeout(0);
+                //Let the eventStream service manage that
+//                session.setSyncDelay(0);
+//                session.setSyncTimeout(0);
```
giomfo: These commented lines should be removed

vector/src/main/java/im/vector/VectorApp.java
```diff
@@ -405,8 +405,9 @@ private void suspendApp() {
         for (MXSession session : sessions) {
             if (session.isAlive()) {
                 session.setIsOnline(false);
-                session.setSyncDelay(pushManager.isBackgroundSyncAllowed() ? pushManager.getBackgroundSyncDelay() : 0);
-                session.setSyncTimeout(pushManager.getBackgroundSyncTimeOut());
+                //This is manage by the event stream service
+//                session.setSyncDelay(pushManager.isBackgroundSyncAllowed() ? pushManager.getBackgroundSyncDelay() : 0);
+//                session.setSyncTimeout(pushManager.getBackgroundSyncTimeOut());
```
giomfo: These commented lines should be removed

vector/src/main/java/im/vector/services/EventStreamServiceX.kt
```diff
@@ -200,7 +208,12 @@ class EventStreamServiceX : VectorService() {
 
         when (action) {
             ACTION_START,
-            ACTION_GO_TO_FOREGROUND ->
+            ACTION_GO_TO_FOREGROUND -> {
+
+                //We are back in foreground, we can sync
+                mSession?.syncDelay = 0
+                mSession?.syncTimeout = 30000
```
giomfo: Where does this value `30000` come from?

akallabeth:
Can anything be done to help getting this pr merged?
Fixing the notification issues with F-Droid builds would really be great.

vector/src/main/java/im/vector/VectorApp.java
BillCarsonFr: done

vector/src/main/java/im/vector/services/EventStreamServiceX.kt
BillCarsonFr: recommended value for riot

UralZima:
there is verbose discussion in https://github.com/vector-im/riot-android/issues/3162
Does this pull request put back background service with background permission as it was in version 0.8.26 (and then this code was removed)? 
Because of this issue I am still using version 0.8.26, but now I can't call ios users anymore because they upgraded to new version.
When to expect a new version with this code included?

akallabeth:
@UralZima this is part of riot `0.9.8` and available with the latest f-droid builds
the sync is better than before but still not on par with the old `0.8.26` behaviour :/

UralZima:
Thanks for answer. The sync is better - it works? Like every 30 sec I'll get notifications with screen off? I am afraid to upgrade, no way to safely downgrade without uninstalling

akallabeth:
@UralZima well, the sync works until your android decides to kill the app, so you need to start it (once a day? think depends on your battery savings)

PR #3161, opened 2019-05-14
Fixing and updating arabic

ghost:


bmarty:
Hi @BoFFire ,

Thanks for the PR

I know we have some difficulties with https://translate.riot.im/, but it is now up.
Are the translation in this PR also added to weblate?

I will merge the PR before the next release to ensure Weblate sync is ok

Benoît

PR #3235, opened 2019-07-30
make Widget/Integration mgr optional (#3224)

### LINKED ISSUE #3224: Make sure there are no ugly edge cases running Riot without an integrations manager
Edge cases we need to manage (display an error or hide the feature):

- integration manager screen
- Sticker picker
- jitsi conf call

iOS issue: https://github.com/vector-im/riot-ios/issues/2518

BillCarsonFr:
Fixes #3224 

Make widget/integration manager optional.
Integration manager is read from preferences with default values to new vector integration server.
Integration (Matrix app / stickers) ui elements are removed when no integration manager is defined

image

image

### DIFF
--- a/vector/src/main/java/im/vector/widgets/WidgetManagerProvider.kt
+++ b/vector/src/main/java/im/vector/widgets/WidgetManagerProvider.kt
@@ -29,17 +29,22 @@ object WidgetManagerProvider {
         if (widgetsManager != null) {
             return widgetsManager
         }
-        val serverURL = PreferencesManager.getIntegrationServerUrl(context)
-        if (serverURL.isNullOrBlank()) return null
+        val uiURl = PreferencesManager.getIntegrationServerUiUrl(context)
+        val apiURL = PreferencesManager.getIntegrationServerApiUrl(context)
+        val jitsiUrl = PreferencesManager.getIntegrationServerJitsiUrl(context)
+        if (uiURl.isNullOrBlank() || apiURL.isNullOrBlank() || jitsiUrl.isNullOrBlank()) return null
         return try {
-            URL(serverURL)
+            //Very basic validity check (well formed url)
+            URL(uiURl)
+            URL(apiURL)
+            URL(jitsiUrl)
             val defaultWhitelist = ArrayList(PreferencesManager.getIntegrationWhiteListedUrl(context))
-            defaultWhitelist.add(0,"$serverURL/api")
+            defaultWhitelist.add(0,apiURL)
 
             val config = IntegrationManagerConfig(
-                    uiUrl = "$serverURL/",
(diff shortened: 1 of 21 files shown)


vector/src/main/res/values/strings.xml
```diff
@@ -809,6 +809,7 @@
     <string name="settings_logged_in">Logged in as</string>
     <string name="settings_home_server">Home Server</string>
     <string name="settings_identity_server">Identity Server</string>
+    <string name="settings_integration_server">Integration Server</string>
```
manuroe: "Integration Manager"

vector/src/main/res/values/config.xml
```diff
@@ -14,8 +14,7 @@
     <string name="bug_report_url" translatable="false">https://riot.im/bugreports/submit</string>
 
     <!-- Widget urls -->
-    <string name="integrations_ui_url" translatable="false">"https://scalar.vector.im/"</string>
-    <string name="integrations_rest_url" translatable="false">"https://scalar.vector.im/api"</string>
+    <string name="integration_server_url" translatable="false">"https://scalar.vector.im"</string>
```
manuroe: I would prefer that we keep 2 app settings for the moment. One for the UI. One for the API.

vector/src/main/java/im/vector/widgets/WidgetManagerProvider.kt
```diff
@@ -0,0 +1,52 @@
...
+        val serverURL = PreferencesManager.getIntegrationServerUrl(context)
+        if (serverURL.isNullOrBlank()) return null
+        return try {
+            URL(serverURL)
+            val defaultWhitelist = ArrayList(PreferencesManager.getIntegrationWhiteListedUrl(context))
+            defaultWhitelist.add(0,"$serverURL/api")
+
+            val config = IntegrationManagerConfig(
+                    uiUrl = "$serverURL/",
+                    apiUrl = "$serverURL/api",
```
manuroe: That seems a bit dangerous atm. `/api` suffix is not guaranteed.
This is something that will change but not before phase:2

vector/src/main/res/values/config.xml
BillCarsonFr: Ok, changed to use 3 settings (added integrations_jitsi_widget_url)

vector/src/main/java/im/vector/widgets/WidgetManagerProvider.kt
BillCarsonFr: changed

vector/src/main/res/values/strings.xml
BillCarsonFr: done

Issue #3121, opened 2019-04-16
Appfdroid build flavour contains google play service libraries


Bubu:
See basically here: https://github.com/jitsi/jitsi-meet/issues/4106

This is blocking any updates of riot-android on F-Droid. The only solution I can think of would be continuing building with the old jitsi sdk until this is solved in some other way.

Can be confirmed with:
```
$ ./gradlew assembleAppfdroidRelease
$  $ANDROID_HOME/tools/bin/apkanalyzer dex packages ./vector/build/outputs/apk/appfdroid/release/vector-appfdroid-release-unsigned.apk | grep com.google.android.gms
```

Bubu:
Building the jitsi SDK without the nonfree components is now possible. It means that riot needs to pull the jitsu SDK from a local maven repo instead then. (And also build master of the SDK, as there isn't a release yet AFAIK.)

Trying that I ran into some weird minsdk version issues as documented here: https://github.com/jitsi/jitsi-meet/issues/4106#issuecomment-493902552

bmarty:
@Bubu just add the missing packages in the manifest like this to solve the compilation issue:

```
<uses-sdk tools:overrideLibrary="org.jitsi.meet.sdk,com.oney.WebRTCModule,co.apptailor.googlesignin,com.BV.LinearGradient,com.calendarevents,com.corbt.keepawake,com.dylanvann.fastimage,com.oblador.vectoricons,com.ocetnik.timer,com.rnimmersive,com.zmxv.RNSound,com.i18n.reactnativei18n,org.webkit.androidjsc,com.reactnativecommunity.asyncstorage,com.reactnativecommunity.webview" />
```

PR #3270, opened 2019-08-23
Temporarily remove Jitsi integration for F-Droid build

bmarty:
- Remove Jitsi integration for the F-Droid version of the application, due to the fact that Jitsi library include non free software (package "co.apptaillor.googlesignin")
More details here: https://github.com/vector-im/riot-android/issues/3121 
- F-Droid applicationId set to "im.vector.app"


bmarty:
Replaced by #3276

PR #3274, opened 2019-08-26
Add and prioritize the new "_matrix/integrations/v1" widget urls

jaywink:
As per MSC1961, add to the whitelisted integrations_widget_urls
the new paths. This allows us to switch Scalar over to use the
new path as default.

Note, the legacy "scalar-staging.riot.im" is these days just a redirect
to scalar-staging.vector.im, so there is no addition for that. It still
needs Riot side whitelisting though for existing widgets.

Refs: https://github.com/matrix-org/matrix-doc/pull/1961
Signed-off-by: Jason Robinson <jasonr@matrix.org>


jaywink:
@manuroe this would be good to push out at some point so that at some point in the future we can switch Scalar over to the MSC1961 API's.

PR #3325, opened 2019-09-25
Try to fix issue 2453

### LINKED ISSUE #2453: Incoming call notification eternal
If someone calls me and I start talking on the PC, and at this point, Android receives the event about the call with a little delay, the notification remains forever.

screenshot_20180720-123348

on develop build G-b2266
Android 7.1.2
room is encrypted.

GammaPi:
Will this fix #2453 ?
I didn't have build environment at hand for testing and not quite familiar with this repository. So I just made those small changes based on my observation:

I found "refreshForegroundNotification" seems to be a function called when synchronising from homeserver. And since mForegroundNotificationState != INCOMING_CALL or CALL_IN_PROGRESS, I just call hideCallNotifications() to remove the notification.

Sorry for not digging deeper into the source code. This bug is really very annoying, please help me to verify whether this fix works. Thanks a lot.


GammaPi:
@bmarty

GammaPi:
After looking through code. I found this fix doesn't work.
But I will provide some details I found in issue 2453

PR #3481, opened 2020-05-27
remove jquery

### LINKED ISSUE #3381: [Security] jquery version has vulnerability
This app includes [jQuery 2.1.4](https://github.com/vector-im/riot-android/blob/71874e68577c29d29764727c16da7ab423e959fb/vector/src/main/assets/html/js/jquery-2.1.4.min.js), which is used in [markdown.html](https://github.com/vector-im/riot-android/blob/71874e68577c29d29764727c16da7ab423e959fb/vector/src/main/assets/html/markdown.html).

This version of jQuery has [known vulnerabilities](https://snyk.io/test/npm/jquery/2.1.4).

Is this actually being used?

Should it be updated / removed as appropriate?

hrj:
Fixes #3381


bmarty:
Hi @hrj , thanks for the PR. I assume you have tested on MiniVector that markdown parsing is still working fine, even on old devices, I'm not a specialist of Web dev :)

turt2live (approved):
It's equivalent code, though I'll leave compatibility checking with you :)

hrj:
Hi @bmarty 

I only tested on Android 9, since I don't have an older device than that. But I did a quick check on compatibility [here](https://caniuse.com/#feat=domcontentloaded). It's a well known site among web devs. The only browser that doesn't support this event is Internet Explorer 8 and below.

MDN also [says](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event) that Android WebView v1+ supports this event.

So I felt it was a safe bet to take.

PR #3512, opened 2020-08-25
Kill Riot-Android (for F-Droid users)

### LINKED ISSUE #3507: Kill F-Droid im.vector.alpha
On the same principle than https://github.com/vector-im/riotX-android/issues/1727, we should properly kill the app `im.vector.alpha` on F-Droid

On F-Droid, a new app has been created, so https://f-droid.org/en/packages/im.vector.alpha/ should be un-installed by user, and they should then install https://f-droid.org/packages/im.vector.app/

A latest update of `im.vector.alpha` on F-Droid should inform the users about that.

We can import the beta popup of RiotX to do so and adapt for F-Droid

Also the description of the legacy app on F-Droid could be updated, maybe we can do this first @Bubu ?

bmarty:
Fixes #3507

If user is already connected, he will sees a gentle reminder each time he launches the app:

image

If user is not connected (or after logout), the popup will be a blocker:

image

This only apply on devices running Android 5.0 and later, because Element Android in only available for such devices.


ganfra (approved):
LGTM

bmarty:
stale PR, let's close it

